PR #18, opened 2018-11-13
TensorFlow 1.12 GPU build script

wdirons:
Builds TensorFlow 1.12 for CUDA 10.0
Uses open source NCCL 2 from github
Uses an ubuntu 18.04 container as the base as that is the only
image with cuda10 that nvidia provides.


wdirons:
@sandipmgiri , @jayfurmanek please review

PR #24, opened 2018-12-05
Update dockerfiles and add publish script

tjakob:
Upstream tensorflow updated their dockerfiles/method of generating them.
This patchset used the new method to generate the dockerfiles, then
made minor changes. All the docker files had at least 1 `ARG` before
the `FROM` in order to select things like ubuntu or cuda versions. This
feature isn't supported until docker version >= 17.05.0, and rhel currently
has docker version 1.13.


tjakob:
@wdirons

tjakob:
@jayfurmanek

tjakob:
Fixed some bugs, and moved a couple things around in the dockerfile to increase CI performance.

PR #31, opened 2019-01-25
Add build script for ONNX

chinhuang007:
Add build script for ONNX. The script is used to set up,
build, and test ONNX and show daily build status for
Linux on ppc64le.

### DIFF
--- a/onnx/setup.sh
+++ b/onnx/setup.sh
@@ -1,0 +1,61 @@
+# ----------------------------------------------------------------------------
+#
+# Package	: ONNX
+# Version	: 1.4
+# Source repo	: https://github.com/onnx/onnx.git
+# Tested on	: ubuntu 18.04
+# Script License: Apache License, Version 2 or later
+# Maintainer	: Chin Huang <chhuang@us.ibm.com>
+#
+# Disclaimer: This script has been tested in non-root mode on given
+# ==========  platform using the mentioned version of the package.
+#             A non-root user must be created with sudo permissions.
+#             The default python version is python 2.7. To build and
+#             test in python3, the environment varible PYTHON_VERSION
+#             must be set to "python3".
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+#!/bin/bash
+
+export NUMCORES=`grep -c ^processor /proc/cpuinfo`
+if [ ! -n "$NUMCORES" ]; then
+  export NUMCORES=`sysctl -n hw.ncpu`
+fi
+echo Using $NUMCORES cores
+
+sudo apt-get update 
+
+sudo apt-get install -y \
+         protobuf-compiler \
+         libprotobuf-dev \
+         cmake 
+
+# Use python3 or python2 venv based on env variable PYTHON_VERSION
+if [ "${PYTHON_VERSION}" = "python3" ]; then
+    sudo apt-get install -y \
+         python3-pip \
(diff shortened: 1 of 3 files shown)


gerrith3:
Priya can you do a final review and pull this if it looks good to you?

seth-priya:
also have the scripts been validated outside of jenkins? Looks like some dependencies / git clone may be missing, could you please test inside a fresh Ubuntu container and update accordingly?

tedhtchang:
@seth-priya We have tested in a fresh container. Here are the additional steps in order to run in a fresh 18.04 container:

#Because OSU Jenkins uses jenkins user to run build step so create a jenkins on the fresh container
$useradd -m -s /bin/bash jenkins
$apt install -y sudo vim
$useradd -m -p time4fun -s /bin/bash jenkins
$visudo
#sudo without password
$jenkins ALL=(ALL) NOPASSWD: ALL
$su - jenkins

#This is taken care by git scm in Jenkins. So without it, here are the additional steps:
$sudo apt update;apt install -y git
$git clone https://github.com/onnx/onnx.git
$cd onnx
$export WORKSPACE=$(pwd)

The rest is running same build steps in https://powerci.osuosl.org/job/onnx-ppc64le-nightly-build/configure to successfully build the job.  
What  would you like us to add to the scripts?

seth-priya:
@tedhtchang thanks!!

I think we should definitely add this part to script, as users more often than not, would first try to use the scripts pushed here outside the Jenkins env.
_#This is taken care by git scm in Jenkins. So without it, here are the additional steps:
$sudo apt update;apt install -y git
$git clone https://github.com/onnx/onnx.git
$cd onnx
$export WORKSPACE=$(pwd)_

As for the first part (installing sudo and creating user), I think it would be sufficient to specify in the script header (as we typically do) that it has been tested with a non-root user
https://github.com/ppc64le/build-scripts/blob/master/ace/ace_rhel_7.2.sh
Please check L1 to L16.

Adding this header helps answer some of the queries that we typically get, upfront...

onnx/setup.sh
```diff
@@ -0,0 +1,55 @@
...
+    sudo apt-get install -y \
+         python-pip 
+fi
+
+python -V
+pip -V
+
+# need to clone source when not running the scripts in Jenkins
+if [ -z "${JENKINS_URL}" ]; then
+    apt install -y git
```
tedhtchang:
```suggestion
    sudo apt install -y git
```

onnx/setup.sh
```diff
@@ -0,0 +1,55 @@
...
+    git clone https://github.com/onnx/onnx.git
+    cd onnx
+    export WORKSPACE=$(pwd)
```
tedhtchang:
```suggestion
        cat <<EOT > .onnxbuild_env
        export WORKSPACE=$(pwd)
EOT
    . .onnxbuild_env
```

onnx/setup.sh
chinhuang007: @tedhtchang Thanks for the review. A new version is pushed with updates.

chinhuang007:
@seth-priya Thanks for the comments. The following are added:

1. code to run a few extra commands if not running in Jenkins
2. headers with general information and a couple of notes for the scripts to run properly

Please let me know if we should add or modify anything.

chinhuang007:
Changed script to use venv for both python2 and python3 therefore should be no dependencies on Jenkins at all. The build log looks clean, https://powerci.osuosl.org/job/onnx-ppc64le-nightly-build/74/. @seth-priya Please review. @tedhtchang Please verify in your fresh container.

PR #41, opened 2019-05-14
Update README.md for FRR

seth-priya:
Add required dependencies


seth-priya:
@avmgithub please review.

PR #43, opened 2019-06-07
Cacheless Docker build performance improvements

wdirons:
To improve performance of docker builds with no local cache, docker pull of the
last pushed image can be used in conjunction with the --cached-from tag in
docker build.

This works until docker build encounters the 'COPY' command in the Dockerfile.
Then it stops using the cache. For this reason we moved the 'COPY' command
in the juypter dockerfiles to the end of the dockerfiles.

Docker pull is only used when the tag is latest. Added the -s option to
skip the pull.

Removed unused 'f:' from getopts and reordered them to match the case statement.

Unrelated:
Added future in the devel images for a recent upstream change.


wdirons:
@tjakob, @seth-priya please review.

@jayfurmanek FYI

seth-priya:
Thanks @wdirons !

PR #44, opened 2019-06-13
Unset TF_NEED_TENSORRT in devel images.

tjakob:
Nvidia released TensorRT with cuda 10.1 support, we are currently using 10.0

@seth-priya @wdirons


seth-priya:
thanks @tjakob @wdirons , merged!

PR #47, opened 2019-06-20
Fix the debian prompt

chinhuang007:
Add -E to sudo to suppress the debian prompt
when installing python-pip. We will also set
the env variable DEBIAN_FRONTEND=noninteractive


chinhuang007:
@seth-priya Can you please review? Thanks!

seth-priya:
Looks, good, thanks!

PR #48, opened 2019-06-24
Fix warning when tensorflow is imported

wdirons:
When importing tensorflow in a python app, it displays the warning:
h5py is running against HDF5 1.10.0 when it was built against 1.8.4, this may cause problems

This warning is false, because it was built against HDF5 1.10.0,
but setup.py was unable to find libhdf5.so to determine the version
number during the build. For this reason we copy libhdf5.so to
/usr/local/lib before the build and remove it after the build.


wdirons:

jayfurmanek:
Looks like I don't have write access (thought I did).
@seth-priya Can you merge?  Thanks!

PR #50, opened 2019-07-08
TensorFlow Serving 1.14.0 updates

wdirons:
The current docker build is failing because the latest version
of numpy can not be installed with python 2.7. Switch to use
python 3.6 instead of python 2.7

Updated bazel to version 0.24.1, with new build paramter to required
for java.

Replaced references to 1.13.0 with 1.14.0


wdirons:

tjakob:
Looks good to me.

seth-priya:
@wdirons looks good, thanks!

PR #51, opened 2019-07-08
Use python3 for tensorflow docker containers

wdirons:
Simple modification to publish.sh to use python3 for all new docker
containers. Basically "latest" and "latest-py3" will be the same
image with different names.

The alternative was to remove all the python 3 or 2 logic, but that
logic might come in handy in the future if we need to support
another python release.

Added a message on what image we are building to the logs to help
with debug.

### DIFF
--- a/tensorflow/Dockerfiles/publish.sh
+++ b/tensorflow/Dockerfiles/publish.sh
@@ -69,6 +69,7 @@ fi
 
 build_push ()
 {
+    echo "Building Docker file with options: Type=$TYPE, Release=$RELEASE, Python_3=$PYTHON_3, JUPYTER=$JUPYTER_ARG"
     if [ $TYPE = "gpu" ]; then
         ARG_SUFFIX="-gpu"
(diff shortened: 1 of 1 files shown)


wdirons:

The end result is two different image names using the same image id.
```
REPOSITORY                                       TAG                     IMAGE ID            CREATED             SIZE
registry-1.docker.io/ibmcom/tensorflow-ppc64le   latest                  0dfbf59aff7a        15 minutes ago      1.16GB
registry-1.docker.io/ibmcom/tensorflow-ppc64le   latest-py3              0dfbf59aff7a        15 minutes ago      1.16GB
registry-1.docker.io/ibmcom/tensorflow-ppc64le   latest-jupyter          2a323382b37f        33 minutes ago      1.37GB
registry-1.docker.io/ibmcom/tensorflow-ppc64le   latest-py3-jupyter      2a323382b37f        33 minutes ago      1.37GB
```

tjakob:
@wdirons Looks good to me. Thank you

PR #53, opened 2019-07-15
Fix tensorflow devel containers

wdirons:
scipy was failing to install in a docker container
first error was: No lapack/blas resources found
second error was no fortran compiler found
third error was cython was not found

Remove enum34 as we no longer support python 2.7

Install cython and numpy first, before the others so the packages
that require them can find them the first time they attempt to compile.
This should save time plus get rid of false error messages in the logs.


wdirons:

PR #54, opened 2019-07-17
Fixes to Docker container exiting on creation.

sarveshtamba:
The image built had issues in it due to which the docker container exited as soon as it was created.
Investigated and found out that 1.) OrientDb http port was not exposed and 2.) Default command to start the OrientDB server were not executed as a part of the Dockerfile.


ghatwala:
LGTM !!!

PR #56, opened 2019-07-22
Latest UBI 8 Dockerfile for OrientDB

sarveshtamba:


sarveshtamba:
> In Docker build steps ...avoid using "ppc64le/orientdb" as there already exists one repo here - https://hub.docker.com/r/ppc64le/orientdb/.

Incoporated the suggested changes. Please review and approve.

PR #63, opened 2019-08-06
UBI based dockerfile for tiDB and LICENSE for cassandra

sarveshtamba:
UBI based dockerfile for tiDB and LICENSE for cassandra


ghatwala:
LGTM

PR #74, opened 2019-09-09
Update for h5py 2 10 0

wdirons:
Update for h5py version 2.10.0

h5py version 2.10.0 was released on September 6th, and it requires
the command pkg-config in order to build from source. In every
docker container except for cpu-ppc64le.Dockerfile we had already
installed pkg-config.

The nice thing about this new version, it is able to build with
libhdf5-dev installed without any hacks to find libhdf5-dev or the
.h file. So these hacks have been removed.

While making this change I also made changes to integrate with latest upstream changes:
    
TensorFlow added new Jupyter notebooks to their docker images in commit:
https://github.com/tensorflow/tensorflow/commit/2feb645aa0ce259f779687add10f9a0e0afeecce

They also added curl to all docker images in commit:
https://github.com/tensorflow/tensorflow/commit/1689e1fe22bf52140f0500a048a03136a466757e


wdirons:

seth-priya:
LGTM!

PR #86, opened 2019-10-18
add patch and buildscript for de.flapdoodle.embed.mongo

shirodkara:
add patch and buildscript for de.flapdoodle.embed.mongo
also contains a patch for de.flapdoodle.embed.process


ghatwala:
LGTM

PR #88, opened 2019-10-21
Added Dockerfile for Redis UBI build

redmark-alt:
Note that the Ubuntu Docker file has `checkout 3.2.9` but I found that the latest is `3.2.13`.

### DIFF
--- a/redis/Dockerfiles/latest_ubi_7/Dockerfile
+++ b/redis/Dockerfiles/latest_ubi_7/Dockerfile
@@ -1,0 +1,22 @@
+FROM registry.access.redhat.com/ubi7/ubi:latest
+
+RUN yum update -y \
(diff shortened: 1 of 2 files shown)


redmark-alt:
I've also provide better updates to the README.md file.

seth-priya:
Looks, good, we may need some additional changes for clearing the certification, but they can be in a subsequent PR, thanks!

redmark-alt:
@here I understand the non-root requirement is needed.  I recommend we process `this` pull request so we don't start stacking commits on top of each other.  It makes the reviews harder to read and understand... unless we `squash commits` which I can do if that is our process.

redmark-alt:
@seth-priya Sorry, I was typing my comment and your were closing the PR.

PR #91, opened 2019-10-25
added build script for spring-boot

shirodkara:
added buildscript for spring-boot

This is a unified buildscript that builds spring-boot along with all its dependencies.


ghatwala:
LGTM

PR #93, opened 2019-10-31
Added Dockerfile for Gerrit UBI build

redmark-alt:
This Dockerfile should be Red Hat ready.  Also, note that I elected not to build the Gerrit 3.0.3 WAR file because Google makes it available for download.

-Ron

Screen Shot 2019-10-31 at 1 24 56 PM


ghatwala:
this looks good to me.

PR #97, opened 2019-11-05
Update libcudnn7 and libnccl2 versions

wdirons:
for TensorFlow Docker containers. When TensorFlow was modified to use
the manylinux 2014 docker image for builds, it was built with cudnn
verison 7.6.3. This causes a runtime error when installed in a container
running 7.4.2. To resolve this error we are updating the latest cudnn
version and updating nccl while we are in the file.


wdirons:

seth-priya:
Thanks @wdirons @jayfurmanek LGTM!

PR #100, opened 2019-11-08
Update cudnn in tensorflow serving

wdirons:
Since TensorFlow is building with CUDNN 7.6.3, need to update
CUDNN in the TensorFlow Serving docker container. Change ENV to ARG
so that in the future the version can be changed during the docker build
using the `--build_arg` command line. Made the same change for Bazel.


wdirons:

To test I pointed Jenkins at my branch and used that to publish the TensorFlow Serving docker containers for 1.15.0 and 2.0.0

PR #124, opened 2020-01-19
Fix scipy for tensorflow serving

wdirons:
scipy is installed automatically as part of installing the tensorflow
wheel file. We need to ensure liblapack-dev and pybind11 are installed
prior to that so that it build successfully.


wdirons:
@ jayfurmanek please review

wdirons:
@jayfurmanek please review

ghatwala:
LGTM .

PR #132, opened 2020-02-07
Move the base url out of dockerfiles and into the publish script

wdirons:
The dockerfiles all have duplicate code to determine the url from which to pull
the tensorflow wheel from. In general there are 4 places to pull the wheel file
from:
cpu release builds
gpu release builds
cpu nightly builds
gpu nightly builds

This logic works fine, but it is better to have it in publish.sh one time, and
then pass the BASE_URL into the dockerfile so it can download and install the
correct wheel.

Sometimes we want to publish a docker image using a URL that isn't one of the 4
builds above. The tool doesn't currently handle that. This PR adds a -u (for url)
option to publish.sh to pass in a URL from which the dockerfile should pull the
tensorflow wheel file from.

This allows for special case builds like LMS and TensorFlow 1.15.X builds to be
published in docker without having to clone the branch and hack the scripts.

### DIFF
--- a/tensorflow/Dockerfiles/cpu-ppc64le-jupyter.Dockerfile
+++ b/tensorflow/Dockerfiles/cpu-ppc64le-jupyter.Dockerfile
@@ -83,27 +83,13 @@ RUN ${PIP} --no-cache-dir install --upgrade \
             h5py \
             pybind11
 
-# Options:
-#   tensorflow
-#   tensorflow-gpu
-#   tf-nightly
-#   tf-nightly-gpu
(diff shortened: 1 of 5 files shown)


wdirons:
Tested by pointing Jenkins at much branch and using it to publish docker images for 1.15.2 and the LMS build.

1.15.2 for GPU is still a problem because the latest dockerfiles use CUDA 10.1 (and not 10.0)

wdirons:
@jayfurmanek please review.

PR #143, opened 2020-02-24
Dockerfiles and build instructions for eclipse-che ubi8 images.

kumarvikas13:


kumarvikas13:
@ghatwala 
1. The original Dockerfile remains unchanged. So either we need an updated contents_sets_epel7.repo or an updated Dockerfile. Since updated contents_sets_epel7.repo follows the structure of the original eclipse-che repository, I think this should be the right approach.
2. Apache license is already added to the main eclipse-che folder. Previous sub-folders do not contain copies of LICENSE files, hence not adding any copy to new sub-folders as well.
3. Tried to squash the commits. But looks like there were other PRs that got merged after I raised the PR. Could you check how this can be resolved.

ghatwala:
suggest closing this PR , and raising a new PR with above ( contents_sets_epel7.repo , apache + single commit) , simplest way i guess :-)

ghatwala:
as suggested above , closing this PR , please raise a new one with just a single commit - @kumarvikas13 similar to this - https://github.com/ppc64le/build-scripts/pull/149

PR #154, opened 2020-03-06
build-script for encoding npm for rhel

sarveshtamba:
build-script for encoding npm for rhel


sarveshtamba:
@ghatwala license file is already present as a part of build script for ubuntu. Refer - https://github.com/sarveshtamba/build-scripts/blob/master/encoding/LICENSE

PR #164, opened 2020-03-17
build-script for markdown-js npm

sarveshtamba:
build-script for markdown-js npm


sarveshtamba:
> @sarveshtamba - please add "apache " license here instead of package specific license ( MIT in this case )

Done, please review and merge if it looks good to you.

PR #178, opened 2020-04-01
Fix ONNX auto gen proto issues

chinhuang007:
Fix ONNX auto gen proto issues by adding parameters that are needed for the latest ONNX version.


chinhuang007:
@ghatwala Glad the build status is back to green for ONNX now, https://github.com/onnx/onnx. Thanks for quick review and approval.

PR #209, opened 2020-05-04
Build script for NPM package - zmq 2.15.3

imamol:
zmq - version 2.15.3
node -v => 12.16.1


imamol:
sure @seth-priya, will keep package version as variable.

PR #213, opened 2020-05-08
Upstream thanos build script and change tektoncd build base dockerfile

snehlatamohite:
Hi ,

I have uploaled thanos build script and change tektoncd build base dockerfile for non-root changes which is required to make pipeline work. 

Thanks,
Snehlata Mohite.


snehlatamohite:
Hi Priya,

Thanks for review. I have added LICENSE. 

Thanks.

PR #270, opened 2020-09-22
Dockerfile for custodia

ezeeyahoo:
Hi,

This is build is working and tested. Please review it.

image

Thanks

### DIFF
--- a/c/custodia/Dockerfile
+++ b/c/custodia/Dockerfile
@@ -1,0 +1,23 @@
+FROM ubuntu:xenial
+ENV TOXENV=py36-noextras
+ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
+RUN \
+    # Install dependencies
+    apt-get update && \
+    apt-get install -y software-properties-common && \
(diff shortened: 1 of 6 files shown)


gerrith3 (commented):
@seth-priya do we usually reference a private github repository for the build, or do we cherry pick the patches needed and pull from the original github URL?  I would think the latter makes most sense, especially when Eshant returns to his regular day job.  Also, Eshant, did you try to get this patch accepted by the custodia community?  I don't remember seeing it, and that is the "ideal" path (although not always possible).

seth-priya:
agreed @gerrith3 - the "ideal" option is to get the patch accepted. If that does not get accepted or get delayed for whatever reasons, we generally create and upload the patch file to this repo. such that it resides next to the build script or dockerfile and can be directly applied - that removes any dependency on private fork.

ezeeyahoo:
Here is merged PR https://github.com/latchset/custodia/pull/249

seth-priya:
ok, cool - so in that case we should be able to just clone the official tree in your dockerfile as opposed to the private fork, I believe?

ezeeyahoo:
I think I have done some mistake. I will open a new PR

PR #272, opened 2020-09-23
Dockerfile for custodia

ezeeyahoo:
Hi,

I have got PR merged (https://github.com/latchset/custodia/pull/249) and this Dockerfile is tested on VM for which screenshots were shared #270 
Please review it.

Thanks

### DIFF
--- a/c/custodia/Dockerfile
+++ b/c/custodia/Dockerfile
@@ -1,0 +1,25 @@
+FROM ubuntu:xenial
+MAINTAINER "Eshant Gupta <guptaeshant@gmail.com>"
+ENV TOXENV=py36-noextras
+ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
+WORKDIR /root
+RUN \
+    # Install dependencies
+    apt-get update && \
+    apt-get install -y software-properties-common && \
+    add-apt-repository ppa:deadsnakes/ppa && \
+    apt-get update  && \
+    apt-get install -yq --no-install-suggests \
+            python3.6 python3.6-dev python3-pip git enchant locales \
+            libffi-dev libssl-dev && \
+    apt-get clean && apt-get autoremove -y && \
+    #locale for en_US
+    locale-gen en_US.UTF-8 && \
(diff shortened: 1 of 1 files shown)


c/custodia/Dockerfile
```diff
@@ -0,0 +1,24 @@
+FROM ubuntu:xenial
+MAINTAINER "Eshant Gupta <guptaeshant@gmail.com>"
+ENV TOXENV=py36-noextras
+ENV LANG='en_US.UTF-8' LANGUAGE='en_US:en' LC_ALL='en_US.UTF-8'
+RUN \
+    # Install dependencies
+    apt-get update && \
```
seth-priya, line 7: we should use the "-y" flag for apt-get update as well as otherwise it can cause the script execution to wait on user input in certain scenarios

c/custodia/Dockerfile
```diff
@@ -0,0 +1,24 @@
...
+    apt-get update  && \
+    apt-get install -yq --no-install-suggests \
+            python3.6 python3.6-dev python3-pip git enchant locales \
+            libffi-dev libssl-dev && \
+    apt-get clean && apt-get autoremove -y && \
+    #locale for en_US
+    locale-gen en_US.UTF-8 && \
+    ln -sf python3.6 /usr/bin/python3 && \
+    # Clone Project
+    git clone https://github.com/latchset/custodia.git /root/custodia && \
```
seth-priya, line 19: let's avoid using hard coded path values, may make more sense to define a WORKDIR or something similar at the top

seth-priya (changes requested):

seth-priya:
Generally looks fine, just a few minor things above and some general comments / thoughts
- wherever possible we should create dockerfiles for stable versions as opposed to master branch. 
- I have not checked but is this an application that provides a service which we can run as a container by exposing ports , providing entrypoint and such or is it more of a library that can be integrated with or which you can build you application?

If its the latter the a build script may make more sense than a dockerfile?

ezeeyahoo:
No issue I will add a build script.

ezeeyahoo (commented):
Done

seth-priya (approved):
LGTM

PR #281, opened 2020-10-08
Build script for CockroachDB v19.2.10 on UBI8

imamol:
- Build script for CockroachDB v19.2.10 on UBI8 and required patches
- Dockerfile


ghatwala (approved):
LGTM

PR #287, opened 2020-10-11
adding buildscript for sysdig

shirodkara:
adding buildscript for sysdig


seth-priya (approved):

seth-priya:
Let's try to stick to a specific version (for example the last release) and making it configurable going forward, given the specific dependencies that are otherwise needed, thanks

PR #296, opened 2020-11-20
Added Dockerfile for elasticsearch v7.8.0

junawaneshivani:


seth-priya (changes requested):
@junawaneshivani we may want to update the comment, especially about the template part to avoid confusion (someone may look for that file in this repo. for instance), lets add an attribution and reference to the source of the Dockerfile and supporting scripts and each of these, also please add maintainer details, thanks!

seth-priya (approved):

PR #297, opened 2020-11-21
Added script for SLES 12 SP5

NishikantThorat:
- Removed libffi and cmake after installations.
- Installed python in home directory
- Created python venv for data agent python specific files.

### DIFF
--- a/d/datadog-agent/README.md
+++ b/d/datadog-agent/README.md
@@ -1,4 +1,4 @@
-# Build datadog-agent
+# Build datadog-agent on RHEL
 
 In order to build datadog-agent, please copy the contents of current directory to a directory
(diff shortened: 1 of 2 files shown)


seth-priya (commented):
Hi Nishikant - I have a couple of questions:
- if there already exists a "go" or "cmake" version on the target system, will the script overwrite it? or will it fail? do we need to handle this explicitly?
- before uninstalling any of the packages, should we check if they were installed by our script (or were already there on the system) and uninstall only if we installed them ? Basically we don't want to remove anything that was already there on the system, correct?

seth-priya (commented):
LGTM

seth-priya (approved):

PR #299, opened 2020-11-24
st2 - checkout specific commit and build st2 cli too

amitsadaphule:


amitsadaphule:
@seth-priya please review. Also, we may need to downgrade MongoDB to 4.0 as per https://github.com/StackStorm/st2/issues/5076. But I'm leaving it as is for now, since only >=4.2.7 for MongoDB is available for ppc64le and build/tests are passing currently with MongoDB 4.4 being used in the script.

seth-priya:
LGTM

seth-priya (approved):

PR #302, opened 2020-12-06
Add UBI 8.2 based build-script for chromium 84.0.4118.0

amitsadaphule:


c/chromium/Chromium_84.0.4118.0_UBI_8.2.sh
```diff
@@ -0,0 +1,180 @@
...
+# Release mode
+is_component_build = false
+is_debug = false
+# Disable broken features
+enable_nacl = false
+treat_warnings_as_errors = false
+enable_dav1d_decoder = false
+# For clang, add the following lines:
+is_clang = true
+clang_base_path = "/root/chromium_build/llvm_build"
```
seth-priya, line 147: I believe this should be relative to CHROMIUM_DIR rather than being hard coded?

c/chromium/Chromium_84.0.4118.0_UBI_8.2.sh
```diff
@@ -0,0 +1,180 @@
...
+cd $CHROMIUM_DIR
+git clone https://chromium.googlesource.com/chromium/tools/depot_tools.git
+
+export DEPOT_TOOLS_UPDATE=0
+export PATH=${CHROMIUM_DIR}/gn/out:$PATH:${CHROMIUM_DIR}/depot_tools
+export VPYTHON_BYPASS="manually managed python not supported by chrome operations"
+export GYP_DEFINES="disable_nacl=1"
+
+# check out LLVM and Clang
+export CLANG_REVISION=375507
```
seth-priya, line 72: how do we map the chromium version to the exact clang version, if there is some commend for doing so, maybe we worth mentioning in a comment for future reference.

c/chromium/Chromium_84.0.4118.0_UBI_8.2.sh
amitsadaphule: right, I'll change that.

c/chromium/Chromium_84.0.4118.0_UBI_8.2.sh
amitsadaphule: Actually, there is a way to map the chromium version to clang version as used [here](https://github.com/leo-lb/electron/blob/7-2-x/ppc64le_build.bash#L85). But since I was referring to [this](https://github.com/ppc64le/build-scripts/blob/master/c/chromium/Chromium_84.0.4119.0_RHEL_7.6.sh#L116) for building chromium, I just used the same version being used there. I'll try updating that to use the command to fetch the version.

amitsadaphule:
@seth-priya please review.

seth-priya (approved):

PR #308, opened 2020-12-22
Fixing some problems found at runtime

oskrdt:
- Using `dumb-init_1.2.2_ppc64le` for Power arch instead of `dumb-init_1.2.2_amd64` for x86 arch
- Moving shebang to the first line of `kibana-docker` script
- Adding step to copy node binaries to final container
- Removing default of `discovery.type=single-node` that prevents to run cluster


seth-priya:
@junawaneshivani

junawaneshivani:
@seth-priya changes look good to me. You can merge this PR.

PR #310, opened 2020-12-24
AddingPowerSupport_CI/Testing

santosh653:
Adding power support ppc64le with Continues Integration/testing so that code remains architecture independent.

This is part of the Ubuntu distribution for ppc64le. This helps us simplify testing later when distributions are re-building and re-releasing.

The build is successful on both arch: amd64/ppc64le, please find the Travis Link below.
https://travis-ci.com/github/santosh653/ginkgo

Please let me know if you need any further details.

Thank You !!


seth-priya:
So @santosh653 there seem to be two travis.yml files that you are trying to add here - one at the top level and one in the travis_yamsl folder? We need just one that will reside in the travis_yamls folders, so please remove the other one and also squash your commits into a single commit.

seth-priya:
@santosh653 the headers are missing /incorrect / incomplete in the arrow travis file and it seems that cfssl upstream repo. already has Power support https://github.com/cloudflare/cfssl/blob/master/.travis.yml, so there is no need to duplicate it here. Finally try to keep it one commit per travis.yml file or clearly state in the PR title/summary as to which all packages are being included.

seth-priya:
Looks like all of these have since been covered in other PRs @santosh653 ? Let me know if that is not the case, but I am closing this for now

PR #313, opened 2020-12-24
Building no-oss verrsion

oskrdt:
I'm changing the build version to use the basic instead of the OSS since this last one lacks several sections/features. You can refer to [this](https://discuss.elastic.co/t/index-management-tab-is-not-visible-after-updating-to-7-9-1/259418) topic that I opened on Elastic portal for more details. 
Using basic version enables several sections as showed on the following image
image


seth-priya:
@junawaneshivani please review

junawaneshivani:
@seth-priya LGTM!

PR #314, opened 2020-12-28
Created doxygen travis.yml file

rajesh-ibm-power:
Created travis.yml file for doxygen package


seth-priya:
@rajesh-ibm-power can we place adhere to the naming convention for the filename as proposed here https://github.com/ppc64le/build-scripts/tree/master/travis-ymls, for example for this one you need to name your file as "doxygen.travis.yml"

PR #315, opened 2020-12-28
Adding power support for this package to support arch independent

asellappen:
Adding power support for this package to support arch independent for web


asellappen:
wrongly created

PR #321, opened 2020-12-28
Create rear.travis.yml

gururajrkatti:
Original .travis.yml was giving parse error. Corrected it. Also added "make install" based on the build script.

### DIFF
--- a/travis-ymls/rear.travis.yml
+++ b/travis-ymls/rear.travis.yml
@@ -1,7 +1,7 @@
 # ----------------------------------------------------------------------------
 #
 # Package             : rear
-# Source Repo         : https://github.com/gururajrkatti/rear
+# Source Repo         : https://github.com/rear/rear
 # Travis Job Link     : https://travis-ci.com/github/gururajrkatti/rear/jobs/465409899
 # Created travis.yml  : No
 # Maintainer          : Gururaj R Katti <Gururaj.Katti@ibm.com>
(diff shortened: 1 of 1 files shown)


seth-priya:
@gururajrkatti do we need this, I see the upstream repo at https://github.com/rear/rear/blob/master/.travis.yml already has ppc64le support?

gururajrkatti:
Hi @seth-priya, thanks for the response and feedback. When I checked the repos, I could not see any travis build running, I tried to trigger it myself, I was seeing parse error. Hence modified it a bit. Also, based on the dockerfile in build script, I added one more step to the travis job.

seth-priya:
Ok, that seems interesting - although I have no particular issues in merging this PR, in this case, I would actually recommend that you go back to the community and *fix* their job, especially if the issue is seen on Intel as well? The install step maybe unnecessary in travis.yml , but again no harm in having it in either

travis-ymls/rear.travis.yml
```diff
@@ -0,0 +1,22 @@
+# ----------------------------------------------------------------------------
+#
+# Package             : rear
+# Source Repo         : https://github.com/gururajrkatti/rear
```
seth-priya, line 4: Please mention the link to the original repo. instead of the forked one.

seth-priya (changes requested):

gururajrkatti:
@seth-priya, community has accepted this change

seth-priya:
Thanks @gururajrkatti for the update and for following through on this! In that case I don't think we need  it here

PR #325, opened 2020-12-29
created overrides.travis.yml for adding power support to the package

asellappen:
Adding power support by CI /testing

### DIFF
--- a/overrides.travis.yml
+++ b/overrides.travis.yml
@@ -1,9 +1,9 @@
 # ----------------------------------------------------------------------------
 #
-# Package             : minipass-pipeline
-# Source Repo         : https://github.com/isaacs/minipass-pipeline
-# Travis Job Link     : https://travis-ci.com/github/asellappen/minipass-pipeline/
-# Created travis.yml  : Yes
+# Package             : overrides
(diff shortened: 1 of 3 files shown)


overrides.travis.yml
```diff
@@ -0,0 +1,19 @@
+# ----------------------------------------------------------------------------
+#
+# Package             : overrides
+# Source Repo         : https://github.com/tlrobinson/overrides
+# Travis Job Link     : https://travis-ci.com/github/asellappen/overrides/
+# Created travis.yml  : No
```
seth-priya, line 6: @asellappen I don't see a travis.yml in the original repo, so shouldn't this be "Yes" or did I miss something?

asellappen: @seth-priya,  Yes ,It's newly created. Pls treat it as 'YES'
I have created newly for another two packages as well and not changed to 'YES' before merging, Now i have done the changes on these package as well ,Apologies for the confusion .

The following are newly created yml and missed to update to Yes in label
1.minipass-collect.travis.yml
2.minipass-pipeline.travis.yml
3.overrides.travis.yml

seth-priya (approved):

PR #349, opened 2020-12-31
Frep_PullRequest

santosh653:
creating frep.travis.yml with proposed changes.


seth-priya:
The upstream frep community already supports ppc64le builds, hence no need to duplicate here. Closing the PR.

PR #454, opened 2021-01-11
Created travis.yml file for package ember-cli

rajesh-ibm-power:
Created travis.yml file for package ember-cli


gururajrkatti (commented):
Is fast_finish useful here ? Rest all LGTM

rajesh-ibm-power:
Fast Finishing is for security propose only, Anyhow its not mandatory but have. fast_finish description goes like below 
If some jobs in the build matrix are allowed to fail, the build won’t be marked as finished until they have completed.

gururajrkatti (approved):
ok

PR #488, opened 2021-01-15
AddPowerSupport_react-html-table-to-excel

santosh653:
Creating react-html-table-to-excel.travis.yml with proposed changes.


Siddhesh-Ghadi:
LGTM @seth-priya

PR #490, opened 2021-01-15
add buildscript for apicurio-registry

shirodkara:
add buildscript for apicurio-registry


seth-priya (commented):
@shirodkara should we update the header to say that this has been tested with "root" mode or install packages using "sudo" perhaps? Can you please check and update?

seth-priya (approved):

PR #500, opened 2021-01-17
Create node-i18n-utilities.yml

nageshlop:
Created new file .travis.yml, added support for ppc64le and verified the changes with build and test.

### DIFF
--- a/travis-ymls/node-i18n-utilities.yml
+++ b/travis-ymls/node-i18n-utilities.yml
@@ -1,0 +1,25 @@
+# ----------------------------------------------------------------------------
+#
+# Package             : node-i18n-utilities
+# Source Repo         : https://github.com/IBM/node-i18n-utilities
+# Travis Job Link     : https://travis-ci.com/github/nageshlop/node-i18n-utilities/builds/213744036
+# Created travis.yml  : Yes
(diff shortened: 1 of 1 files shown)


travis-ymls/node-i18n-utilities.yml
```diff
@@ -0,0 +1,24 @@
...
+# Script License      : Apache 2.0
+#
+# ----------------------------------------------------------------------------
+dist: xenial
+arch:
+   - amd64
+   - ppc64le  
+language: node_js
+node_js:
+   - "10"
```
seth-priya, line 18: I see that the repo is not maintained so the latest node versions might not work but any particular reason to select version 10?

nageshlop: @seth-priya I referred the previous travis builds and as the project is related to javascript added nodejs 10 and tested. Also verified with nodejs 15 version its working fine. Let me know how to go about this.
Build Log: https://travis-ci.com/github/nageshlop/node-i18n-utilities/builds/213742618

seth-priya: I believe they have moved to another CI and that is why their travis builds may not be updated? In any case, would suggest adding the latest versions as well.

nageshlop:
@seth-priya ready for merge.

seth-priya (approved):

PR #502, opened 2021-01-18
Create readline.travis.yml

gururajrkatti:


seth-priya (commented):
don't we have any unit test cases to be executed to be here?

gururajrkatti:
> don't we have any unit test cases to be executed to be here?

README file doesn't mention about any specific tests. Build docker file in build_scripts includes "make install" in addition to the steps used in travis. Should it be included in travis ?

seth-priya:
can you check if there is a "check" or "test" target in the Makefile?

gururajrkatti:
> can you check if there is a "check" or "test" target in the Makefile?

"check" target is there. But it is empty. "make check" doesn't do anything. No "test" target.

seth-priya:
ok thanks

seth-priya (approved):

PR #520, opened 2021-01-19
Added yml to support power on the package markdown

asellappen:


gururajrkatti (commented):
What is the reason for explicitly specifying the dist as xenial for some versions and no explicit mention for some versions ? Rest all LGTM

asellappen:
some versions of python having issue in build if it is other than xenial dist .

gururajrkatti (approved):
ok

PR #528, opened 2021-01-20
Add power support in travis for proj

nagesh4193:
I have added power support for proj.

### DIFF
--- a/travis-ymls/proj.travis.yml
+++ b/travis-ymls/proj.travis.yml
@@ -1,0 +1,94 @@
+# ----------------------------------------------------------------------------
+#
+# Package             : proj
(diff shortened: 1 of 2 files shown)


gururajrkatti (commented):
Is one of the jobs added "publish_doc" doing anything ? Rest all LGTM

nagesh4193:
> Is one of the jobs added "publish_doc" doing anything ? Rest all LGTM

For intel version AMD64, it is publish doc. The time taken to run the job is less.
For arm64 and s390x, it is linux generic. But the time taken to run these is more.
So, I have added power architecture support with both types of builds as both are passing.

PR #562, opened 2021-01-25
AddingPowerSupport_flask-restful

santosh653:
Creating flask-restful.travis.yml with proposed changes.

### DIFF
--- a/travis-ymls/flask-restful.travis.yml
+++ b/travis-ymls/flask-restful.travis.yml
@@ -1,0 +1,160 @@
+# ----------------------------------------------------------------------------
+#
+# Package             : flask-restful
+# Source Repo         : https://github.com/flask-restful/flask-restful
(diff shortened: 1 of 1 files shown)


gururajrkatti (changes requested):
Should we not retain 2.7 as that is the only python supported in python2 ?

santosh653:
@gururajrkatti I think we are ok if the code runs fine on python3 versions, there were failures on python2.7, I confirm on this from my previous interactions with Priya.

santosh653:
@gururajrkatti Can you pl review?

gururajrkatti:
Are the same failures found on intel also ?

santosh653:
@gururajrkatti , Yes Guru, there were similar failures on both Intel & power. That's why we are good here I think.

gururajrkatti (approved):
ok

santosh653 (commented):
@seth-priya Can you please merge this PR now?

PR #581, opened 2021-01-28
Adding power support to package jnr-constants.travis.yml

asellappen:


gururajrkatti (changes requested):
I see a build failure here ? Also is it required to add multiple versions of JDK for ppc64le ? As per logs, internally its using JDK 11 only.

asellappen:
> I see a build failure here ? Also is it required to add multiple versions of JDK for ppc64le ? As per logs, internally its using JDK 11 only.

Failure is due to unavailable version of jdk for amd64 , and as per my understanding we are not taking care of amd64 issues ,hence i left since I have just modified existing yml. Now i have removed that to make it clean run.

And also added only jdk11 for ppc64le.

gururajrkatti (changes requested):
Oraclejdk11 not supported on ppc64le. You can remove it. For the amd issue, you can raise an issue/PR with the community.

asellappen:
> Oraclejdk11 not supported on ppc64le. You can remove it. For the amd issue, you can raise an issue/PR with the community.

ppc64le - removed Oraclejdk11 .
amd64 issue - unavailable support version of oraclejdk8 for amd64 (Expected feature release number in range of 9 to 17, but got: 8), since this is also unsupported version ,so it is no point to raise the issue to community.

pls review now

gururajrkatti (approved):
ok

PR #598, opened 2021-02-01
Updated buildscript for apache-flink

bivasda1:
Build and E2E tests passed successfully.


seth-priya (commented):
@bivasda1 can we add a new script for RHEL 8 instead of modifying the old one?

bivasda1:
@seth-priya Ok I will add new script for RHEL 8.

PR #621, opened 2021-02-04
Created react-intl-redux.travis.yml

dthadi3:
created react-intl-redux.travis.yml and added support ppc64le and amd64.


gururajrkatti (approved):
LGTM

PR #622, opened 2021-02-04
Created travis.yml for package dex

rajesh-ibm-power:
Created travis.yml for package dex


gururajrkatti (commented):
I think other CI using "make testall" which looks similar to this. LGTM

PR #627, opened 2021-02-04
Create minimal-json.travis.yml

nageshlop:
Added support for ppc64le.

### DIFF
--- a/travis-ymls/minimal-json.travis.yml
+++ b/travis-ymls/minimal-json.travis.yml
@@ -1,0 +1,19 @@
+# ----------------------------------------------------------------------------
+#
+# Package             : minimal-json
+# Source Repo         : https://github.com/ralfstx/minimal-json
+# Travis Job Link     : https://travis-ci.com/github/nageshlop/minimal-json/builds/217195379
+# Created travis.yml  : No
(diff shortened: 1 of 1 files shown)


gururajrkatti (changes requested):
What is the reason for using jdk11 for amd and jdk8 for ppc ? Also you don't include it and then exclude it, you can have one include job for amd and one for ppc

nageshlop:
@gururajrkatti Its not openjdk11 its oraclejdk11 which is part of existing travis.yml and runs on amd64 architecture and the same is excluded for ppc64le as its not supported. Hence, I have added openjdk8 for ppc64le which is supported.

gururajrkatti:
You don't need to include it and then exclude it

nageshlop:
@gururajrkatti oraclejdk11 is added for amd64 and excluded for ppc64le as it is not supported.

nageshlop:
@gururajrkatti can you review, updated the travis.yml file to remove exclude part and verified the changes with build and test.

gururajrkatti (approved):
ok

nageshlop:
@seth-priya  ready to merge.

PR #636, opened 2021-02-05
Adding power support  pillow

asellappen:


gururajrkatti (commented):
Isn't the install.sh (for installing dependencies) to be executed before build.sh ?

asellappen:
> Isn't the install.sh (for installing dependencies) to be executed before build.sh ?

Done,Pl. check now

gururajrkatti (approved):
ok

PR #649, opened 2021-02-09
Arumugam work branch packaging

asellappen:


gururajrkatti (commented):
1. Is it not possible to run single command each in install and script similar to intel ?
2. lint job doesn't seem to run anything. Is there any issue there ?

asellappen:
> 1. Is it not possible to run single command each in install and script similar to intel ?
> 2. lint job doesn't seem to run anything. Is there any issue there ?

1. not possible , Pl check previous runs for the same package and i have already tried many ways to port like intel, It does not work.
2. yes. It looks nox Sessions lint not support in both cpu arch ,I have tried with different versions of python and not working.

gururajrkatti (approved):
ok

PR #671, opened 2021-02-11
Added castor.travis.yml

sanjaymsh:
Hi,

I have added castor.travis.yml.  I have replaced oraclejdk7 and 8 by openjdk9 & 10 And added arch as amd64 and ppc64le. The travis build log can be tracked here https://travis-ci.com/github/sanjaymsh/castor/builds/216765641 . Please review and merge it.

Regards,
Sanjay

### DIFF
--- a/travis-ymls/castor.travis.yml
+++ b/travis-ymls/castor.travis.yml
@@ -1,0 +1,16 @@
+# ----------------------------------------------------------------------------
+# Package             : castor
+# Source Repo         : https://github.com/castor-data-binding/castor
+# Travis Job Link     : https://travis-ci.com/github/sanjaymsh/castor/builds/216765641
+# Created travis.yml  : No
+# Maintainer          : Sanjay Mishra <snjkmr32@gmail.com>
+# Script License      : Apache License, Version 2 or later
+# ----------------------------------------------------------------------------
+language: java
+arch:
+  - amd64
+  - ppc64le
+jdk:
+  - openjdk9
+  - openjdk10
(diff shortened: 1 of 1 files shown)


travis-ymls/castor.travis.yml
```diff
@@ -0,0 +1,16 @@
...
+# Created travis.yml  : No
+# Maintainer          : Sanjay Mishra <snjkmr32@gmail.com>
+# Script License      : Apache License, Version 2 or later
+# ----------------------------------------------------------------------------
+language: java
+arch:
+  - amd64
+  - ppc64le
+jdk:
+  - openjdk9
```
gururajrkatti, line 14: On amd64, currently it uses oraclejdk, should we not leave it as it is instead of changing it to openjdk ? For ppc64le, it may be better to use jdk 11

gururajrkatti (changes requested):

travis-ymls/castor.travis.yml
sanjaymsh: @gururajrkatti , It failing for amd64 if i am keeping as it is with below errors: 

install-jdk.sh 2020-06-02
Expected feature release number in range of 9 to 17, but got: 8
The command "~/bin/install-jdk.sh --target "/home/travis/oraclejdk8" --workspace "/home/travis/.cache/install-jdk" --feature "8" --license "BCL"" failed and exited with 3 during .

gururajrkatti (approved):
ok

PR #676, opened 2021-02-12
Create rdma-core.travis.yml

ddeka2910:


gururajrkatti (commented):
I don't see any tests being run. Can you pl. try "make test" or "make check"

ddeka2910:
The commands "make test" and "make check" doesn't trigger any test.
However the "build.sh" script seems to be performing some tests as can be seen in the log. Could you please verify the same and confirm.

gururajrkatti:
There is "tests" folder in the repos. Can you pl. check if any of them are being run ?

ddeka2910:
@gururajrkatti Tried pytest. But it is giving errors.

gururajrkatti:
We need to check what are the errors

seth-priya:
The testing itself has not happened as per the comments above, so substantial rework would be needed if someone else needs to pick this up, closing the PR as a fresh one would make more sense

PR #677, opened 2021-02-12
Create registry-url.travis.yml

dthadi3:
Created registry-url.travis.yml and added support for power and amd64.


gururajrkatti (approved):
LGTM

PR #706, opened 2021-02-17
Includes tcl/tk capability

sebacea:


gerrith3 (commented):
I see no issues here.  Just one question - does this make the image noticeably larger?  Also, once this is committed, I assume we need to re-publish the rstudio container to dockerhub with a new tag.  Once this is committed, @seth-priya can see if we can find someone to do that, although because of our workload in may take a couple of weeks.

gerrith3 (approved):
LGTM

sebacea:
Hi, do you have more updates to support R 4.2 (updating to ubuntu focal) and a more recent version of studio server. I did some proofs without success

seth-priya:
@sebacea unfortunately no, we have not worked on rstudio recently, we typically upgrade only when we get a specific request rather than proactively keeping it upto date and we have not received one recently. If you can create an issue with the problems you are running into we might be able to help provide some help cc @gerrith3

PR #709, opened 2021-02-18
Create demjson.travis.yml

nageshlop:
Created travis.yml file and added support for ppc64le. Have verified whether the package supports higher versions of python and it didn't go through.


gururajrkatti (commented):
Any specific reason for not using latest python/python 3.x ?

nageshlop:
@gururajrkatti  Have verified whether the package supports higher versions of python but it fails.

gururajrkatti (approved):
ok

PR #714, opened 2021-02-19
Add Power support for Kie Cloud Operator

nagesh4193:


seth-priya (changes requested):
Please move the already checked in file from the top level folder to travis.ymls instead of adding a new copy.

seth-priya:
addressed with https://github.com/ppc64le/build-scripts/commit/9505c76467ed9650770a113cceb41909fe004f0f

PR #754, opened 2021-03-05
Add Power support for tinkerpop travis file

nagesh4193:
Added Power support for tinkerpop travis file. I have excluded the .net version as it is not supported for ppc64le. I have commented it in the travis file as well.


gururajrkatti (approved):
LGTM

PR #763, opened 2021-03-09
Create codeready-workspaces.travis.yml

nageshlop:
Created package specific travis file and added support for ppc64le.

### DIFF
--- a/travis-ymls/codeready-workspaces.travis.yml
+++ b/travis-ymls/codeready-workspaces.travis.yml
@@ -1,0 +1,18 @@
+# ----------------------------------------------------------------------------
+#
+# Package             : codeready-workspaces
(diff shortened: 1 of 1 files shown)


gururajrkatti (changes requested):
I think "mvn clean install" itself is running the test also. Do we need "mvn test" again ? Rest all LGTM

nageshlop:
@gururajrkatti yes, just included if in future additional tests are added.

gururajrkatti:
> @gururajrkatti yes, just included if in future additional tests are added.

I think its repeating the same tests ? Is it not duplication ?

nageshlop:
Ok I will remove "mvn test".

nageshlop:
Updated travis.yml file and squashed.

gururajrkatti (approved):
ok

PR #766, opened 2021-03-10
add requestretry_rhel_7.6.sh trim-newlines_rhel_7.6.sh build script

sudeepraj2:


r/requestretry/requestretry_rhel_7.6.sh
```diff
@@ -0,0 +1,56 @@
...
+then
+        nvm install v12.19.1
+fi
+
+nvm alias default v12.19.1
+
+#Installing nodejs v12.19.1 by using nvm
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.32.0/install.sh | bash
+source ~/.nvm/nvm.sh
+nvm install 12.19.1 && node -v
```
amitsadaphule, line 46: node is already at the start of the script [here](https://github.com/ppc64le/build-scripts/pull/766/files#diff-cc3b2315683e35b60103853214ebac26965b0454e8033c5333aef13d566c8a61R38). I think this re-installation needs to be removed.

t/trim-newlines/trim-newlines_rhel_7.6.sh
```diff
@@ -0,0 +1,56 @@
...
+then
+        nvm install v12.19.1
+fi
+
+nvm alias default v12.19.1
+
+#Installing nodejs v12.19.1 by using nvm
+curl -o- https://raw.githubusercontent.com/creationix/nvm/v0.32.0/install.sh | bash
+source ~/.nvm/nvm.sh
+nvm install 12.19.1 && node -v
```
amitsadaphule, line 46: node is already at the start of the script [here](https://github.com/ppc64le/build-scripts/pull/766/files#diff-89554a0d975960009a6701e52efb7c4846e25f16bc131285aaf25413f495b1d6R38). I think this re-installation needs to be removed.

amitsadaphule (changes requested):

r/requestretry/requestretry_rhel_7.6.sh
sudeepraj2: Updated the build script.

t/trim-newlines/trim-newlines_rhel_7.6.sh
sudeepraj2: Updated the build script.

amitsadaphule (approved):
LGTM

PR #774, opened 2021-03-16
Fix labels and README for cassandra dockerfile

amitsadaphule:


amitsadaphule:
@seth-priya please review.

c/cassandra/Dockerfiles/3.11.x-ubi8/Dockerfile
```diff
@@ -4,11 +4,13 @@
 
 FROM registry.access.redhat.com/ubi8/ubi:latest
 
+ARG VERSION=3.11.x
```
seth-priya, line 7: wouldn't it be better to specify the exact version here, instead of saying "x"? or is there a specific reason for that?

amitsadaphule: The reason behind x there is that depends upon the build number "x" in the cassandra rpm version @ https://downloads.apache.org/cassandra/redhat/311x/, which will keep changing. So, user has to specify the correct version as argument while building the image. If we hard-code "x" to a specific value, it may not match to the current build number of rpm at a later point.

seth-priya (approved):

PR #787, opened 2021-03-23
Create observatorium.travis.yml

asellappen:


travis-ymls/observatorium.travis.yml
```diff
@@ -0,0 +1,25 @@
...
+ - ppc64le
+  
+go:
+ - 1.15
+ - 1.14
+
+script:
+ - export PATH=${PATH}:`go env GOPATH`/bin
+ - go build -v ./...
+ - go test -v ./...
```
gururajrkatti, line 25: Apart from these unit tests, there are integrations tests configured through test/integration.sh. Is it possible to run them ?

asellappen: i don t think the scripts  under test folder integration.sh,load.sh and run-local.sh are effectively used and working ,I tried  integration.sh and running for long ,not returning. 

I checked the current ci circleci  pipelines at   https://app.circleci.com/pipelines/github/observatorium/observatorium
and the test integration.sh is not  being done .I have seen only  build, unit test and lint checks are done there.

now i have included lint check also for amd64 alone(ppc64le not supporting due missing dep) in our travis build.

pl review.

gururajrkatti: In this circleci links I see some additional tests ending with "Logs Read/Write tests: OK  "

gururajrkatti:
what is the change from previous submission ? Are we able to run additional tests (logs ending with "Logs Read/Write tests: OK " seen in circleci) ?

asellappen (commented):
> what is the change from previous submission ? Are we able to run additional tests (logs ending with "Logs Read/Write tests: OK " seen in circleci) ?

make test --always-make; was missing when we comapr circleci ,so included this test and it works only for amd64 and you can find the content "Logs Read/Write tests: OK " at the end of log file for mad64.


it does not work for ppc since failing to download Loki for ppc64le for running "make test --always-make" , not found suitable version at https://github.com/grafana/loki/releases for ppc64le.

gururajrkatti (approved):
ok

PR #791, opened 2021-03-24
Create dockerize.travis.yml

asellappen:


gururajrkatti (commented):
go test doesn't seem to run any tests.. Can we run "make" to verify it ?

asellappen:
> go test doesn't seem to run any tests.. Can we run "make" to verify it ?

Yes, There is no test scripts or test files under this package and this is the reason go test not run any tests.

And also it will be the same case even we run make also, If we check make file for this this package then there is no rule specified for testing or verifying & checking.

gururajrkatti (approved):
ok

PR #811, opened 2021-04-05
Create javax-inject_ubuntu_20.04.sh

gururajrkatti:
build script for javax-inject

### DIFF
--- a/j/javax-inject/javax-inject_ubuntu_20.04.sh
+++ b/j/javax-inject/javax-inject_ubuntu_20.04.sh
@@ -1,0 +1,31 @@
+# ----------------------------------------------------------------------------
+#
+# Package	: javax-inject
+# Version	: 4.0.0 
+# Source repo	: https://github.com/javax-inject/javax-inject 
+# Tested on	: ubuntu_20.04
+# Script License: Apache License, Version 2 or later
+# Maintainer	: Gururaj R Katti <Gururaj.Katti@ibm.com>
+#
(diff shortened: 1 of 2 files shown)


j/javax-inject/javax-inject_ubuntu_20.04.sh
```diff
@@ -0,0 +1,31 @@
...
+#!/bin/bash
+
+sudo apt-get update
+sudo apt-get install openjdk-8-jdk wget git -y
+export JAVA_HOME=/usr/lib/jvm/java-8-openjdk-ppc64el/
+
+wget http://www-eu.apache.org/dist/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
+tar xzvf apache-maven-3.6.3-bin.tar.gz
+export PATH=$PATH:`pwd`/apache-maven-3.6.3/bin
+git clone https://github.com/javax-inject/javax-inject
```
seth-priya, line 26: I believe the reason we are going with top-of-tree/master is that this project does not have a current release or tag that we can use, is that so?

gururajrkatti: Yes @seth-priya it doesn't have any release tags

seth-priya:
@gururajrkatti also, we need the Apache 2.0 license file to go along with each script

gururajrkatti:
@seth-priya I have added the license file. Pl. check

seth-priya (approved):

PR #824, opened 2021-04-14
Create jackson-jaxrs-json-provider.travis.yml

genisysram:


gururajrkatti (commented):
Is it not the duplicate of #813 ? Do we need two travis.yml files here ?

PR #825, opened 2021-04-14
enable activeMQ build

afsanjar:


gerrith3 (changes requested):
Okay - would would it take to base this on ubi8 instead of centos:7?  Since 7 is old and centos is kind of going away, this would future proof us for a while.

I didn't see a LICENCE file - each build script should be licensed under Apache 2.0.

I defer to Priya for any other comments.

a/activemq/Dockerfiles/latest_centos_7/build/Dockerfile
```diff
@@ -0,0 +1,7 @@
+FROM ppc64le/centos:7
```
seth-priya, line 1: agree with Gerrit on using ubi 8 instead of centos:7 unless of course there is an explicit need for centos:7 and I believe the image should come with a multiarch manifest, so we may not need the "ppc64le/" part explicitly

a/activemq/Dockerfiles/latest_centos_7/build/build.sh
```diff
@@ -0,0 +1,7 @@
+git clone https://github.com/apache/activemq
```
seth-priya, line 1: we generally keep the build scripts in the top level package folder which would be https://github.com/ppc64le/build-scripts/tree/master/a/activemq in this case and also have a specific header for each script like https://github.com/ppc64le/build-scripts/blob/master/a/aaargh/aaargh_rhel_7.3.sh#L1 to https://github.com/ppc64le/build-scripts/blob/master/a/aaargh/aaargh_rhel_7.3.sh#L16 which lists down the package, distro version, Maintainer name etc. which makes them consistent and easier to share with other teams etc. The header also has the script license which we set as "Apache 2.0"

a/activemq/Dockerfiles/latest_centos_7/run/Dockerfile
```diff
@@ -0,0 +1,18 @@
+FROM ppc64le/centos:7
```
seth-priya, line 1: same comment as earlier about use of centos:7 and "ppc64le/"

seth-priya (changes requested):

a/activemq/Dockerfiles/latest_centos_7/build/Dockerfile
afsanjar: Regarding selecting centos7, I agree with @gerrith3 and @seth-priya. However, ActiveMQ builds/Run requires java-7 https://activemq.apache.org/version-5-getting-started.html#installation-procedure-for-unix. I'll try ubi8, not sure if it supports JAVA 7.

afsanjar: regarding the license file, there was already a one at the root https://github.com/ppc64le/build-scripts/tree/master/a/activemq

a/activemq/Dockerfiles/latest_centos_7/run/Dockerfile
afsanjar: okay

a/activemq/Dockerfiles/latest_centos_7/build/build.sh
afsanjar: Thanks, will do

gerrith3 (approved):
Well, that looks much better, and much simpler to me!

The only minor thing that I would love to be able to add is the ability to specify the Version on the command line - by default build 5.17.1 but if someone wants to try another version, it is simply an argument, e.g. 5.17.2 and following.  Might break for 5.18 or not, but then we can get a new issue raised.  However, that is a  "would be nice" - not sure if Priya has a boilerplate for that, since we probably have only used it occasionally.

Oh, and is there a plan to publish this to dockerhub, quay,io, icr, gcr, red hat marketplace, etc.?  Since this is a commonly used dependency it would be nice to get it published at least one high visible place with appropriate tags.

Letting Priya do a final review/approval (or change request) but I'm good with it as it stands.

seth-priya:
yes, we have not been very consistent with the version bit, but here is a recent example https://github.com/ppc64le/build-scripts/blob/master/j/JTransforms/JTransforms_ubuntu_16.04.sh#L22
essentially just have a default one and give the user to enter his own. But we should probably do more checking as to the validity etc, else it will just fallback to master

Approving this for now

seth-priya (approved):

PR #858, opened 2021-04-28
Add build script for 3scale_toolbox

Abhijit-Mane:
This PR adds build/test steps for **3scale_toolbox** (https://github.com/3scale/3scale_toolbox) on ppc64le architecture.
It also does Unit & Integration tests. 

Signed-off-by: Abhijit Mane <abhijman@in.ibm.com>

### DIFF
--- a/3/3scale_toolbox/build.sh
+++ b/3/3scale_toolbox/build.sh
@@ -1,0 +1,50 @@
+# ----------------------------------------------------------------------------
+# Package       : 3scale_toolbox 
+# Version       : v0.18.1
+# Source repo   : https://github.com/3scale/3scale_toolbox
+# Tested on     : RHEL_8.2
+# Script License: Apache License, Version 2 or later
+# Maintainer    : Abhijit Mane <abhijman@in.ibm.com>
+#
+# Disclaimer: This script has been tested in non-root mode on given platform using
+#             the mentioned version of the package. It may not work as expected 
+#             with newer versions of the package and/or distribution.
+#             In such case, please contact "Maintainer" of this script.
+# ----------------------------------------------------------------------------
+
+
+#!/bin/bash
+
+# pre-reqs
+yum install -y gem redhat-rpm-config ruby-devel zlib-devel.ppc64le git make cmake 
+yum install -y zlib-devel.ppc64le gcc.ppc64le gcc-c++.ppc64le
+
+# clone and checkout last stable release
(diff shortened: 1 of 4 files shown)


3/3scale_toolbox/build.sh
```diff
@@ -0,0 +1,50 @@
+# ----------------------------------------------------------------------------
+# Package       : 3scale_toolbox 
+# Version       : v0.18.1
```
krishvoor, line 3: Can you kindly point to this Tag/Branch 3scale-2.10.0-GA?
https://github.com/3scale/3scale_toolbox/archive/refs/tags/3scale-2.10.0-GA.tar.gz

3/3scale_toolbox/build.sh
```diff
@@ -0,0 +1,50 @@
...
+# pre-reqs
+yum install -y gem redhat-rpm-config ruby-devel zlib-devel.ppc64le git make cmake 
+yum install -y zlib-devel.ppc64le gcc.ppc64le gcc-c++.ppc64le
+
+# clone and checkout last stable release
+git clone https://github.com/3scale/3scale_toolbox
+cd 3scale_toolbox
+
+RELEASE_TAG=v0.18.1
+git checkout $RELEASE_TAG
```
krishvoor, line 27: This can be removed, if we are downloading this directly
https://github.com/3scale/3scale_toolbox/archive/refs/tags/3scale-2.10.0-GA.tar.gz

krishvoor (commented):
Minor nits, the rest of the content /lgtm

3/3scale_toolbox/build.sh
gerrith3: I would recommend setting up the version as a build argument where possible - and the default should be the most recent one we ported.  That way if there are minor changes, the latest versions may build; if not, over time if if/switch/case statement to build correctly for different versions would b good.  Even in OSPat, we label things as a specific version or newer, knowing that most of the time, the newer version will probably work.  And we are going to go through a number of versions before we ship this, I"m sure.

gerrith3: BTW, I don't often see us downloading the tar-gz file - it might be slightly faster, especially for really large projects, but I think most to a git checkout, and it might be easier to use a random branch name (release or interim release) using git clone and a checkout of the specific release tag.  But beyond that, I don't see much difference between the two approaches.

PR #862, opened 2021-04-29
Add build-scripts for 3scale-istio-adapter and keyclock-containers

krishvoor:
This PR can be triggered in multiple ways

1) For build on a master, specific branch
2) For Build & Testing on a master, specific branch

Signed-off-by: Krishna Harsha Voora <krishvoor@in.ibm.com>

### DIFF
--- a/3/3scale-istio-adapter/Dockerfiles/Dockerfile
+++ b/3/3scale-istio-adapter/Dockerfiles/Dockerfile
@@ -1,0 +1,11 @@
+FROM registry.access.redhat.com/ubi8/ubi:latest
+
+MAINTAINER Krishna Harsha Voora <krishvoor@in.ibm.com>
+
+
+# Install Prerequisites
+RUN yum install -y git make gcc wget
+
+RUN cd 
+
+CMD ["bash"]
--- a/3/3scale-istio-adapter/Dockerfiles/build.sh
+++ b/3/3scale-istio-adapter/Dockerfiles/build.sh
@@ -16,7 +16,7 @@
 
 if [ -z $1 ] || [ "$1" == "laststablerelease" ]
 then
-	RELEASE_TAG=code-cleanups
+	RELEASE_TAG=v2.0.0.1
 else
 	RELEASE_TAG=$1
 fi
@@ -73,3 +73,7 @@ fi
 # Copy Binaries
 cp _output/3scale-istio-adapter /ws/
 cp _output/3scale-config-gen /ws/
+
+# Clean Up
+cd $HOME
+rm -rf $GOPATH/
--- a/3/3scale-istio-adapter/Dockerfiles/build.sh
+++ b/3/3scale-istio-adapter/Dockerfiles/build.sh
@@ -1,0 +1,75 @@
+# ----------------------------------------------------------------------------
+# Package       : 3scale-istio-adapter 
+# Version       : code-cleanups
+# Source repo   : https://github.com/3scale/3scale-istio-adapter
+# Tested on     : RHEL_8.2
+# Script License: Apache License, Version 2 or later
+# Maintainer    : Krishna Harsha Voora
+#
+# Disclaimer: This script has been tested in non-root mode on given platform using
+#             the mentioned version of the package. It may not work as expected 
+#             with newer versions of the package and/or distribution.
+#             In such case, please contact "Maintainer" of this script.
+# ----------------------------------------------------------------------------
+
+#!/bin/bash
+
+if [ -z $1 ] || [ "$1" == "laststablerelease" ]
+then
+	RELEASE_TAG=code-cleanups
+else
+	RELEASE_TAG=$1
+fi
+
+echo "RELEASE_TAG = $RELEASE_TAG"
+
+# Download/Setup Golang 1.11
+cd /usr/local/
+wget https://golang.org/dl/go1.11.13.linux-ppc64le.tar.gz
+tar -zxf go1.11.13.linux-ppc64le.tar.gz
+rm -f go1.11.13.linux-ppc64le.tar.gz
+
+
+# Update PATH Variables
+export GOPATH=$HOME/ibm/
+export PATH=$GOPATH/bin/:$PATH:/usr/local/go/bin/
+
+# Repository Path
+mkdir -p $GOPATH/bin/
+mkdir -p $GOPATH/src/
+mkdir -p $GOPATH/pkg/
+mkdir -p $GOPATH/src/github.com/3scale/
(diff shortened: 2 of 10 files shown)


krishvoor:
@gerrith3 - Created a new PR.
When tested with the latest golang, the UT breaks. So resorted to [CI Documentation(s)](https://github.com/3scale/3scale-istio-adapter/blob/master/.circleci/config.yml#L8) golang requirement.

3/3scale-istio-adapter/Dockerfiles/build.sh
```diff
@@ -0,0 +1,75 @@
...
+	# Unit Tests
+	make unit
+
+	# Integration Tests
+	make integration
+fi
+
+# Copy Binaries
+cp _output/3scale-istio-adapter /ws/
+cp _output/3scale-config-gen /ws/
```
Abhijit-Mane, line 75: Maybe a good idea to cleanup after copying binaries. Other than that LGTM.
# cleanup
cd ..
rm -rf 3scale-istio-adapter

Abhijit-Mane (commented):
Not sure if we should add a copy of the LICENSE, maybe under: -

build-scripts/tree/master/3/3scale-istio-adapter
OR 
build-scripts/tree/master/3

Abhijit-Mane (commented):
Please ignore, accidentally clicked "Re-request review"

3/3scale-istio-adapter/Dockerfiles/build.sh
krishvoor: > Maybe a good idea to cleanup after copying binaries. Other than that LGTM.

I think we may not need this, as we are deleting the container after copying the binaries, so I have not included the step, I can in case it is required.

```docker run --rm -v `pwd`:/ws 3scale-istio-adapter:ibm bash -l -c "cd /ws; ./build.sh <rel_tag> <runtest> | tee output.log"```

**Comment by krishvoor** [comment]
https://github.com/ppc64le/build-scripts/pull/862#pullrequestreview-648908481
Good catch, I missed it. I have added the LICENSE now.

File: 3/3scale-istio-adapter/Dockerfiles/build.sh (hunk 1 above)
Comment (Abhijit-Mane) [reply 3/3scale-istio-adapter/Dockerfiles/build.sh:75]: container is cleaned up but the cloned repo dir would persist on the host at `pwd`/3scale-istio-adapter.
Up to you but maybe we can leave it in this case as repo may not be huge.
Otherwise LGTM.

**Review by Abhijit-Mane — COMMENTED** [review]
Looks good to me.

File: 3/3scale-istio-adapter/Dockerfiles/build.sh [hunk 2]
```diff
@@ -0,0 +1,75 @@
...
+#
+# Disclaimer: This script has been tested in non-root mode on given platform using
+#             the mentioned version of the package. It may not work as expected 
+#             with newer versions of the package and/or distribution.
+#             In such case, please contact "Maintainer" of this script.
+# ----------------------------------------------------------------------------
+
+#!/bin/bash
+
+if [ -z $1 ] || [ "$1" == "laststablerelease" ]
```
gerrith3, line 17: I don't get the point of the $1 == laststablerelease  ?  It looks like if someone specifies that word they get that stable old branch, maybe the one you worked on?  I think you are trying to indicate the most recently tested, which is good, but we should probably standardize on that name.  I guess the name of that branch (code-cleanups?) bugs me a little since it is a pre-release version, and ideally this would get updated to the next release which support Power.  Has "laststablerelease" been used anywhere else?

3/3scale-istio-adapter/Dockerfiles/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM registry.access.redhat.com/ubi8/ubi:latest
+
+MAINTAINER Krishna Harsha Voora <krishvoor@in.ibm.com>
+
+
+# Install Prerequisites
+RUN yum install -y git make gcc wget
```
gerrith3, line 7: If you've installed these before the build, you probably want to remove them after the build to keep the container smaller.

3/3scale-istio-adapter/Dockerfiles/Dockerfile
```diff
@@ -0,0 +1,11 @@
...
+
+RUN cd 
+
```
gerrith3, line 10: I'm confused on this one - this looks like it is just going to run a bash script, not the 3scalie-istio-adapter?  Also, I see you reference installing go in the build script but the dockerfile doesn't indicate that, nor does it looks like it does anything to create this container?

3/3scale-istio-adapter/Dockerfiles/build.sh
gerrith3: I'm not quite sure where this ended up, but we should always clean up anything we don't need in the container at the end, same thing with the build script.  Maybe don't delete if the build fails, so someone can easily debug without having to reinstall dependencies, but if it succeeds, clean up all but the necessary binaries or equivalent.

gerrith3 (changes requested):
I think overall the Dockerfile needs more contents, and a couple of other nits.

3/3scale-istio-adapter/Dockerfiles/Dockerfile
Abhijit-Mane: @gerrith3 Please refer README.md that talks about how the container and build script need to be used. Hopefully, that will make it clear.

3/3scale-istio-adapter/Dockerfiles/build.sh
Abhijit-Mane: Maybe 'lasttestedrelease' can be used which indicates the last release/branch we tested at the time the script was written.
And this will be the default behavior. A specific release can be built by explicitly passing it as a build argument. 
README provides a few script usage examples.

Abhijit-Mane (approved):

3/3scale-istio-adapter/Dockerfiles/Dockerfile
krishvoor: The "life" of the container is short-spanned, and its purpose being is spinning it up for a quick build and test.
I will add the changes.

3/3scale-istio-adapter/Dockerfiles/build.sh
krishvoor: The requested changes are now in place.

Abhijit-Mane (approved):
Looks good to me.

PR #863, opened 2021-04-29
AMQStreams: adding build and test support for strimzi-kafka-oauth

afsanjar:


s/strimzi-kafka-oauth/redhat_ubi8/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM registry.access.redhat.com/ubi8/ubi
```
krishvoor, line 1: Hi Amir, 
Can we make use of this Container Image instead `registry.access.redhat.com/ubi8/openjdk-8` ?

afsanjar: I wish we could, it is corrupted for both x86 and Power. I tried it last week :(

s/strimzi-kafka-oauth/redhat_ubi8/build.sh
```diff
@@ -0,0 +1,49 @@
...
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+#!/bin/bash
+if [ -z $1 ] 
+then
+	# Master branch
+	BRANCH=""
```
Abhijit-Mane, line 22: If no branch/tag is explicitly specified, don't we plan to clone the last stable release instead of master (which isn't always guaranteed to be stable)?

Abhijit-Mane (commented):
Looks good to me.

s/strimzi-kafka-oauth/redhat_ubi8/build.sh
afsanjar: We need a master branch for nightly CI/CD.

Abhijit-Mane (approved):
All comments addressed and looks okay to me.

PR #864, opened 2021-04-29
AMQStreams: enabling cruise-control

afsanjar:
Required by AMQStream

### DIFF
--- a/c/cruise-control/redhat_ubi8/Dockerfile
+++ b/c/cruise-control/redhat_ubi8/Dockerfile
@@ -1,0 +1,5 @@
+FROM registry.access.redhat.com/ubi8/ubi
+RUN yum install vim java-1.8.0-openjdk-devel maven git -y
+env JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk
+env JRE_HOME=/usr/lib/jvm/java-1.8.0-openjdk/jre
+CMD ["bash"]
--- a/c/cruise-control/redhat_ubi8/build.sh
+++ b/c/cruise-control/redhat_ubi8/build.sh
@@ -1,0 +1,35 @@
+# ----------------------------------------------------------------------------
+#
+# Package       : cruise-control
+# Version       : master (default)
+# Source repo   : https://github.com/linkedin/cruise-control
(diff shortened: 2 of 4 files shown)


krishvoor (commented):
Hi Amir,
Can you kindly add the Apache2 LICENSE File, please?

c/cruise-control/redhat_ubi8/build.sh
```diff
@@ -0,0 +1,35 @@
...
+
+#!/bin/bash
+if [ -z $1 ] 
+then
+	# Master branch
+	BRANCH=""
+else
+	BRANCH="--branch "$1
+fi
+BUILD_DIR=""cruise-control
```
Abhijit-Mane, line 26: Is this a type and you instead meant  BUILD_DIR=**"cruise-control"**  ?

afsanjar: good catch, done

Abhijit-Mane (commented):
Looks good to me.

c/cruise-control/redhat_ubi8/Dockerfile
```diff
@@ -0,0 +1,5 @@
+FROM registry.access.redhat.com/ubi8/ubi
```
gerrith3, line 1: Why not use a base of ubi8/openjdk-8 here instead of installing it separately?  Should be faster.

gerrith3 (changes requested):
Lmk if the updated ubi base image doesn't work, but I think that should simplify things just a tiny bit.  It probably also will protect you from CVE exposure by not fixing the OpenJDK maintenance version and it should be handled by Red Hat updated that ubi base image.

c/cruise-control/redhat_ubi8/Dockerfile
afsanjar: Correct @gerrith3, but as you might recall I reported two weeks ago that ubi8/openjdk-8 image was not working, it had JBOS as the user and not root.

Abhijit-Mane (approved):
All changes incorporated, looks okay to me.

seth-priya (approved):

PR #871, opened 2021-05-05
Build Script for Istio Operator (maistra-2.0.1)

anantpednekar-ibm:


seth-priya:
@pratham-m please review

i/istio-operator/istio-operator.sh
```diff
@@ -0,0 +1,45 @@
...
+#Clone repo
+git clone https://github.com/maistra/istio-operator.git
+cd istio-operator
+git checkout tags/maistra-2.0.1 
+
+#Build the Package
+make
+
+#Test the PAckage
+#make test
```
pratham-m, line 44: Why is the test execution commented?

pratham-m: Is that is because of test failures? If yes, then add a comment that some tests are fail on both x86 and ppc64le. And add link (if any) to any open issue on the repository

anantpednekar-ibm: Test execution commented out as some tests fail on  both x86 and ppc64le.
We cannot raise an issue on the repository as its is disabled.

pratham-m: ok. in that case everything else LGTM

pratham-m:
@anantpednekar-ibm kindly squash the commits.

pratham-m:
@seth-priya LGTM

amitsadaphule:
Hey, it seems that the LICENSE file was missed. Can you please add?

pratham-m:
good catch @amitsadaphule
@anantpednekar-ibm kindly add the License file to the `istio-operator`  folder. You can copy it from the root folder of the repo.

PR #884, opened 2021-05-11
Adding build script for ppc64le on zjsonpatch

asellappen:
Adding build script for ppc64le on zjsonpatch

### DIFF
--- a/z/zjsonpatch/zjsonpatch_ubuntu_18.04.sh
+++ b/z/zjsonpatch/zjsonpatch_ubuntu_18.04.sh
@@ -1,0 +1,97 @@
+# ------------------------------------------------------------------------------------------
+# Package          : zjsonpatch
+# Version          : 0.4.11
+# Source repo      : https://github.com/flipkart-incubator/zjsonpatch
+# Tested on        : ubuntu_18.04
+# Passing Arguments: 1.Version of package
+# Script License   : Apache License, Version 2 or later
+# Maintainer       : Arumugam N S <asellappen@yahoo.com> / Priya Seth<sethp@us.ibm.com>
+#
+# Disclaimer       : This script has been tested in non-root mode on given
+# ==========   platform using the mentioned version of the package.It may not
+#              work as expected with newer versions of the package and/or distribution.
+#              In such case, please contact "Maintainer" of this script.
+#
+# ------------------------------------------------------------------------------------------
+
+#!/bin/bash
+
+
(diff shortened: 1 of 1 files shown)


z/zjsonpatch/zjsonpatch_ubuntu_18.04.sh
```diff
@@ -0,0 +1,97 @@
...
+grep -v '<goal>sign</goal>' pom.xml>/tmp/pom.xml
+mv /tmp/pom.xml pom.xml
+
+#Build and test zjsonpatch
+sudo mvn clean install -DskipTests=true -B -V
+ret=$?
+if [ $ret -eq 0 ] ; then
+  echo  "Done build ......"
+else
+  echo  "Failed build ....."
```
gururajrkatti, line 87: Should we not exit here instead of continuing with test. Also in the RHEL script they are using an option "gpg.skip", is it not required here ?

gururajrkatti: @asellappen I see that this pkg already available in OSPAT. Is this script still required ?

asellappen: > @asellappen I see that this pkg already available in OSPAT. Is this script still required ?

@gururajrkatti,  i am not sure about OSPAT, Can you elaborate this ?

asellappen: > Should we not exit here instead of continuing with test. Also in the RHEL script they are using an option "gpg.skip", is it not required here ?

not sure about RHEL,may be that particular task is failing and then they made it to skip. But our case it is getting suc for all tasks.

gururajrkatti: OSPAT is the repository https://power-developer.mybluemix.net/packages/ which lists out the packages present in IBM repos.  Anyways, you can submit this ubuntu script after adding the "exit"

gururajrkatti (changes requested):

z/zjsonpatch/zjsonpatch_ubuntu_18.04.sh
asellappen: Thanks for sharing details on OSPAT ,script for ubuntu is not there.
Added exit and tested now, pls review.

gururajrkatti (approved):
ok

PR #888, opened 2021-05-13
Create connector-configuration-factory_ubuntu_20.04.sh

nageshlop:
Created build script and verified the changes on local VM.

### DIFF
--- a/c/connector-configuration-factory/LICENSE
+++ b/c/connector-configuration-factory/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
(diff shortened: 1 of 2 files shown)


gururajrkatti (changes requested):
Nagesh, I see that the script is exactly same as configuration-file-store-connector__ubuntu_20.04.sh.
If it is same script for multiple jars, we can have one single script right ?

nageshlop:
@gururajrkatti I am fine to make multiple packages having one single script. As of now I could see around 382 packages getting build from egeria main git repository. If that is the case, I will check in the sheet which are all the packages getting build from Egeria and mark them as pass if build is successful. In some cases I found the build is failing for which I can mark them as failed. Could you confirm this is ok to proceed?

nageshlop:
@gururajrkatti  closing all related and dependent packages build script. As discussed will be creating single script to take argument as package name, and run the build and update the respective jar file details.

PR #890, opened 2021-05-13
add buildscript for breakpad

shirodkara:


seth-priya (commented):
can we take version as a parameter instead of hard coding it? and also can we mention the exact RHEL8 version on which it was tested?

seth-priya (approved):

PR #897, opened 2021-05-18
Changes for Psad scripts as per last call with Bhargavi

NishikantThorat:
Adding changes for,

TODO things for Nishikant,
-	Update install and build script for UBI 8.3 container, so user can run build and install script in two different container smoothly.
-	Investigate, why fpm is behaving differently for UBI 8.3 and RHEL 8.2, while generating rpm for psad.


seth-priya:
@sarveshtamba @amitsadaphule please review

p/psad/psad.sh
```diff
@@ -0,0 +1,70 @@
...
+                        ;;
+                c) echo "Building and installing Psad ..."
+			Build
+			Install
+                        ;;
+		h) echo "Usage: "
+                   echo "-b    Only builds psad."
+                   echo "-i    Only installs psad. Kindly ensure, you will have RPM for psad, before using this option."
+                   echo "-c    Build and install psad."
+                   echo "-h    Print usage."
```
amitsadaphule, line 67: Please fix the indentation.

p/psad/psad_RHEL_8.2.sh
```diff
@@ -0,0 +1,46 @@
...
+#
+# Disclaimer    : This script has been tested in root mode on given
+# ==========      platform using the mentioned version of the package.
+#                 It may not work as expected with newer versions of the
+#                 package and/or distribution. In such case, please
+#                 contact "Maintainer" of this script.
+# ----------------------------------------------------------------------------
+yum install -y git wget vim make gcc initscripts chkconfig iptables iproute net-tools  ruby-devel gcc make rpm-build rubygems whois
+gem install --no-document --bindir=/usr/local/bin  fpm
+#sudo gem install --no-document fpm
```
amitsadaphule, line 20: Please remove un-necessary/irrelevant comments.

amitsadaphule (commented):
1. Does `psad_rhel_8.3_install.sh` have a relevance anymore?
- If yes, it'll need fixes too e.g. [./psad_rhel_8.3.sh](https://github.com/NishikantThorat/build-scripts/blob/psad/p/psad/psad_rhel_8.3_install.sh#L11) does not exist anymore.
- If not, let's remove it.
2. If we're able to generate the expected rpm in correct format using `psad_RHEL_8.2.sh` which can be installed and executed properly, can we maybe add an install script for that one?

p/psad/psad_RHEL_8.2.sh
NishikantThorat: Done

p/psad/psad.sh
NishikantThorat: >     1. Does `psad_rhel_8.3_install.sh` have a relevance anymore?

This file is renamed. 
 
>     1. If we're able to generate the expected rpm in correct format using `psad_RHEL_8.2.sh` which can be installed and executed properly, can we maybe add an install script for that one?

We had given install script to Bhargavi as build and install machines were not same and we were trying to install rpm generated from UBI 8.3 on RHEL 8.2. For RHEL 8.2, build and install box will remain same and I was able to install rpm using "rpm -ivh <pkg>" command and able to start psad without any issue on "129.40.81.88". (I think, we dont have any other  box with RHEL 8.2)

Also, have corrected indentation.

amitsadaphule: New commit `124739d deleting install script` took care of #1.

amitsadaphule:
LGTM

amitsadaphule (approved):

PR #899, opened 2021-05-19
adding openj9 JVM

afsanjar:
AMQStreams
Kafka comes with basic OAuth2 support in the form of SASL based authentication module which provides client-server retrieval, exchange and validation of access token used as credentials. For real world usage, extensions have to be provided in the form of JAAS callback handlers which is what Strimzi Kafka OAuth does.


krishvoor (approved):
/lgtm

Abhijit-Mane (approved):
lgtm

PR #915, opened 2021-05-26
Build script for graylog2-server

ManirajDeivendran:


g/graylog2-server/graylog_master_rhel_8.3.sh
```diff
@@ -0,0 +1,66 @@
...
+#
+# --------------------------------------------------------------------------------------
+# Note: Graylog2-server application had a dependency with testcontainers.
+# ====  Testcontainers communicate with other containers through docker.sock.
+#       Use options (--network="host" & -v /var/run/docker.sock:/var/run/docker.sock) if
+#       running the graylog2-server using docker container environment.
+# --------------------------------------------------------------------------------------
+#!/bin/bash
+
+VERSION=master
```
seth-priya, line 21: I believe we should be building/testing this once against the latest stable release as well and if that works, specify that version here instead, because "master" will keep changing with time and it will be difficult to keep track of which code / which point we actually did the validation.

seth-priya (changes requested):

g/graylog2-server/graylog_master_rhel_8.3.sh
ManirajDeivendran: Checkout master with specific commit id 20204a7979569aef116a1bdf30f4b9df6c52bc4d due to release 3.3.13/4.0.7 is unstable.

seth-priya:
looks good, please squash the commits

ManirajDeivendran:
Squashed the commits.

PR #924, opened 2021-06-01
Build script for spring-boot 2.5.0

mdafsanhossain:
This PR is for adding the build script of spring-boot 2.5.0 on ubi8.3.


ghatwala:
@mdafsanhossain  - can you check confirm how is this build script so different to existing  (`spring-boot.sh`) https://github.com/ppc64le/build-scripts/blob/master/s/spring-boot? we see a lots of details in there , can you cross verify ? cc : @NishikantThorat

ghatwala:
Also i dont see any tests being executed in your script in this PR , we should be running the default tests and validate them on power too.

ghatwala (approved):
LGTM

PR #948, opened 2021-06-17
Adding patch and build support for Protoc 3.5.1

afsanjar:
Adding a patch, build script, and install instruction for "protoc-3.5.1-1-linux-ppcle_64.exe for power."
Required by Apache HBase, Zeppelin, and camel-example.

Note, the build script is not version-specific but was not tested with other versions.

### DIFF
--- a/p/protoc/ubuntu-20.04/Dockerfile
+++ b/p/protoc/ubuntu-20.04/Dockerfile
@@ -1,0 +1,11 @@
+FROM ubuntu:20.04
+
+MAINTAINER "Amir Sanjar <amir.sanjar@ibm.com>
+ENV DEBIAN_FRONTEND noninteractive
+
+RUN  apt-get update -y \
+     && apt-get install -y git autoconf libtool automake g++ make curl unzip \
+     && apt install openjdk-11-jdk maven -y
+
+CMD ["bash"]
+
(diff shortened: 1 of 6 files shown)


p/protoc/ubuntu-20.04/protoc-3.5.1.patch
```diff
@@ -0,0 +1,65 @@
...
+diff --git a/protoc-artifacts/pom.xml b/protoc-artifacts/pom.xml
+index 0f9dd9f..4a0892e 100644
+--- a/protoc-artifacts/pom.xml
++++ b/protoc-artifacts/pom.xml
+@@ -37,7 +37,7 @@
+       <extension>
+         <groupId>kr.motd.maven</groupId>
+         <artifactId>os-maven-plugin</artifactId>
+-        <version>1.2.3.Final</version>
++        <version>1.5.0.Final</version>
```
krishvoor, line 62: Just a question, any plans to make this change upstream?

p/protoc/ubuntu-20.04/Dockerfile
```diff
@@ -0,0 +1,11 @@
+FROM ubuntu:20.04
```
krishvoor, line 1: Can we instead make use of `adoptopenjdk`
Or is this a requirement from upstream/ downstream to be Ubuntu?

krishvoor (commented):
Minor nit, otherwise lgtm.

p/protoc/ubuntu-20.04/protoc-3.5.1.patch
afsanjar: Good question, it is already upstream :) I just backported it to version 3.5.1

p/protoc/ubuntu-20.04/Dockerfile
afsanjar: Apache HBase and Bigtop don't use adoptopenjdk.

krishvoor (approved):
/ltgm

PR #974, opened 2021-07-13
Added buildscript for wildfly

csrividya:
This PR consists  of build script for Wildfly and License file.

### DIFF
--- a/w/wildfly/wildfly_ubi_8.3.sh
+++ b/w/wildfly/wildfly_ubi_8.3.sh
@@ -1,0 +1,53 @@
+# ----------------------------------------------------------------------------
+#
+# Package        : Wildfly
+# Version        : 24.0.0.Final
+# Source repo    : https://github.com/wildfly/wildfly
+# Tested on      : ubi:8.3
+# Script License : GNU Lesser General Public License Version 2.1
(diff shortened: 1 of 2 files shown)


w/wildfly/wildfly_ubi_8.3.sh
```diff
@@ -0,0 +1,44 @@
...
+dnf -y install wget
+wget https://www-eu.apache.org/dist/maven/maven-3/3.6.3/binaries/apache-maven-3.6.3-bin.tar.gz
+tar -zxvf apache-maven-3.6.3-bin.tar.gz
+mv apache-maven-3.6.3 /opt/maven
+export JAVA_HOME=/usr/lib/jvm/java-11-openjdk-11.0.11.0.9-2.el8_4.ppc64le
+export M2_HOME=/opt/maven
+export PATH=${M2_HOME}/bin:${PATH}
+
+#Clone wildfly repo
+tagName=24.0.0.Final
```
seth-priya, line 37: can we accept this as a parameter to the build script so that it can be controlled by the user?

seth-priya (changes requested):

seth-priya (approved):

PR #988, opened 2021-08-05
Updated script to built chromium tag 84.0.4118.0

ManirajDeivendran:


c/chromium/Chromium_84.0.4118.0_UBI_8.2.sh
```diff
@@ -4,9 +4,9 @@
 # Package	: chromium
 # Version	: 84.0.4118.0
 # Source repo	: https://chromium.googlesource.com/chromium/src.git
-# Tested on	: UBI 8.2
+# Tested on	: UBI 8.2/UBI 8.4
```
amitsadaphule, line 7: If we've not tested the updated script on UBI 8.2, let's remove it.

amitsadaphule (changes requested):
Except for the one comment related to base OS, rest looks good to me.

c/chromium/Chromium_84.0.4118.0_UBI_8.2.sh
ManirajDeivendran: Removed "UBI 8.2" from comment section. Also renamed the file from Chromium_84.0.4118.0_UBI_8.2.sh to Chromium_84.0.4118.0_UBI.sh

amitsadaphule (approved):
LGTM

PR #997, opened 2021-08-18
MongoDB Dockerfile for ppc64le

ManirajDeivendran:
Dockerfile to create mongo client image version 4.4.

### DIFF
--- a/m/mongodb/Dockerfiles/rhel_8.3/Dockerfile
+++ b/m/mongodb/Dockerfiles/rhel_8.3/Dockerfile
@@ -1,0 +1,135 @@
+FROM ppc64le/ubuntu:20.04
+
+# add our user and group first to make sure their IDs get assigned consistently, regardless of whatever dependencies get added
+RUN groupadd -r mongodb && useradd -r -g mongodb mongodb
+
+RUN set -eux; \
+	apt-get update; \
+	apt-get install -y --no-install-recommends \
+		ca-certificates \
+		jq \
+		numactl \
+	; \
+	if ! command -v ps > /dev/null; then \
+		apt-get install -y --no-install-recommends procps; \
+	fi; \
+	rm -rf /var/lib/apt/lists/*
+
+# grab gosu for easy step-down from root (https://github.com/tianon/gosu/releases)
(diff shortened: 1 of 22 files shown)


m/mongodb/Dockerfiles/rhel_8.3/Dockerfile
```diff
@@ -0,0 +1,135 @@
+FROM ppc64le/ubuntu:20.04
```
seth-priya, line 1: the base image here is ubuntu:20.04 as opposed to RHEL or UBI, which is fine for now but we might have to think of creating an image with UBI base in future. For now, please correct the folder structure. This file should reside in an ubuntu_20.04 folder under Dockerfiles.

m/mongodb/Dockerfiles/rhel_8.3/docker-entrypoint.sh
```diff
@@ -0,0 +1,383 @@
+#!/bin/bash
+set -Eeuo pipefail
```
seth-priya, line 2: General comment on this and the dockerfile - has this been adapted from a similar version already available for other architectures elsewhere? If so, we should have a clear attribution as a comment at the top of each file mentioning its origin

seth-priya (changes requested):

m/mongodb/Dockerfiles/rhel_8.3/Dockerfile
ManirajDeivendran: Understood. Resolved.

m/mongodb/Dockerfiles/rhel_8.3/docker-entrypoint.sh
ManirajDeivendran: Add comment with origin information.

ManirajDeivendran:
Fixed review comments.

seth-priya (approved):
Changes look good, please squash the commits

ManirajDeivendran:
Created new PR https://github.com/ppc64le/build-scripts/pull/1000.

PR #1003, opened 2021-08-26
added build script and patch file for the version v7.8.1

sachinkakatkar:
Tested for the components Elasticsearch-core, aggs-matrix-stat-client and lang-mustache-client


seth-priya:
@kandarpamalipeddi I think we should be fine as it goes into the same elasticsearch folder where the other scripts reside, hence no need to add a license file?

kandarpamalipeddi (approved):

PR #1056, opened 2021-09-27
Build script for psycopg2ciff ppc64le ubi8.3

sumitd2:


snehlatamohite (changes requested):
Kindly change psycopg2_ubi8_3.bash  to .sh to keep uniform naming convention.

PR #1063, opened 2021-09-29
Build and Install Script for fontawesome-free 5.1.0 ubi 8.3

sumitd2:


snehlatamohite:
@sumitd2  - Whether we tried to build required version of fontawesome-free from source and ran tests if yes then kindly share tests execution logs ?

PR #1087, opened 2021-10-07
Build script for lego golang package

ezeeyahoo:
Hi

Please review and provide your feeback.

Thanks

### DIFF
--- a/l/lego/lego_ubi8.3.sh
+++ b/l/lego/lego_ubi8.3.sh
@@ -1,0 +1,85 @@
+#!/bin/env bash
+# ----------------------------------------------------------------------------
+#
+# Package        : Lego
+# Version        : 3.5.0 - latest
+# Source repo    : https://github.com/go-acme/lego
+# Tested on      : Ubi 8.3 + Go latest version (v1.7.1)
+# Script License : Apache 2.0
+# Maintainer     : Eshant Gupta <eshant.gupta@ibm.com>; Priya Seth <sethp@us.ibm.com>
+#
+# Disclaimer: This script has been tested both in root and user mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# Usage: ./lego_ubi_8.3.sh
+#                  OR
+#        ./lego_ubi_8.3.sh v3.5.0
+# ----------------------------------------------------------------------------
+
+set_env=false
+
+# Get sudo access
+[ "$UID" -eq 0 ] || exec sudo bash "$0" "$@"
+
+function usage() {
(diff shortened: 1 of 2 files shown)


l/lego/lego_ubi8.3.sh
```diff
@@ -0,0 +1,78 @@
...
+
+    echo "Create the skeleton for your local users go directory"
+    mkdir -p go/{bin,pkg,src}
+
+    echo "Setting up GOPATH"
+    echo "export GOPATH=$(pwd)/go" >> .profile && source .profile
+
+    echo "Setting PATH to include golang binaries"
+    echo "export PATH='$PATH':$(pwd)/golang/go/bin:$GOPATH/bin" >> .profile && source .profile
+
```
amitsadaphule, line 66: I think there's no need to add things to `.profile` etc and complicate the `go` installation steps. Please refer to [this](https://github.com/ppc64le/build-scripts/commit/49ae79e15cd863475fb705a204293982fe3e95bb#diff-4ba2744d2c13d2a179b0ee4edbb7a701da7697a167fd13f1ff8c9084d4e44353R23-R29)

l/lego/lego_ubi8.3.sh
```diff
@@ -0,0 +1,78 @@
...
+fi
+
+# install lego package
+GO111MODULE="off" go get github.com/go-acme/lego
+cd $GOPATH/src/github.com/go-acme/lego
+go get -tags $TAGS -t ./...
+make test && make build
+[ $? != 0 ] && exit 1
```
amitsadaphule, line 74: `set -eu` can be used at the beginning of the script to get rid of these command status checks.

amitsadaphule (changes requested):

l/lego/lego_ubi8.3.sh
```diff
@@ -0,0 +1,85 @@
...
+# ----------------------------------------------------------------------------
+#
+# Package        : Lego
+# Version        : 3.5.0 - latest
+# Source repo    : https://github.com/go-acme/lego
+# Tested on      : Ubi 8.3 + Go latest version (v1.7.1)
+# Script License : Apache 2.0
+# Maintainer     : Eshant Gupta <eshant.gupta@ibm.com>; Priya Seth <sethp@us.ibm.com>
+#
+# Disclaimer: This script has been tested both in root and user mode on given
```
amitsadaphule, line 11: To keep it simple, we can simply have this script run as root user. Being a _go_ package, it's not necessary to have a capability to run as non-root.

l/lego/lego_ubi8.3.sh
amitsadaphule: We can simply get rid of these writes to .profile/.bashrc/... We're running the script in a UBI container and it is ok to assume that the script will be run in a fresh container.

Siddhesh-Ghadi:
This PR can be closed in favor of #1221.

PR #1088, opened 2021-10-08
Updated build script template for Python, node, PHP, GO, Ruby and Java

Ryan-Dmello:


templates/build_script_go.sh
```diff
@@ -15,66 +15,115 @@
...
+    mkdir -p /home/tester/go/src /home/tester/go/bin /home/tester/go/pkg
 
 export PATH=$PATH:/bin/go/bin
 export GOPATH=/home/tester/go
 
-OS_NAME=`python3 -c "os_file_data=open('/etc/os-release').readlines();os_info = [i.replace('PRETTY_NAME=','').strip() for i in os_file_data if i.startswith('PRETTY_NAME')];print(os_info[0])"`
+mkdir -p output
 
-mkdir -p /home/tester/$PACKAGE_NAME
+OS_NAME=`python3 -c "os_file_data=open('/etc/os-release').readlines();os_info = [i.replace('PRETTY_NAME=','').strip() for i in os_file_data if i.startswith('PRETTY_NAME')];print(os_info[0])"`
```
kandarpamalipeddi, line 34: Not a compulsory change, just a suggestion:
Can we replace this with :
`cat /etc/os-release | grep PRETTY_NAME | cut -d= -f2`
I think, we are doing same using  python.
As we are using shell script,  I guess above command might be more readable.

templates/build_script_python.sh
```diff
@@ -30,65 +32,65 @@ pip3 install -r /home/tester/output/requirements.txt
 pip3 freeze > /home/tester/output/available_packages.txt
 
 PACKAGE_INFO=`cat available_packages.txt | grep $PACKAGE_NAME`
```
kandarpamalipeddi, line 34: Should it be :
```
PACKAGE_INFO=`cat /home/tester/output/available_packages.txt | grep $PACKAGE_NAME`
```
instead of
```
PACKAGE_INFO=`cat available_packages.txt | grep $PACKAGE_NAME`
```

Ryan-Dmello: Made suggested change

Ryan-Dmello:
Also I've incorporated the other 2 suggestion from above

PR #1120, opened 2021-10-21
Adding build script for scala-library. Default version '2.13.6'.

gitssinghal:
tested version 2.12.12 and 2.12.4 from event stream list

### DIFF
--- a/s/scala-library/scala-library_ubi8.sh
+++ b/s/scala-library/scala-library_ubi8.sh
@@ -1,0 +1,40 @@
+PACKAGE_VERSION=2.13.6
+
+echo "Usage: $0 [-v <PACKAGE_VERSION>]"
+echo "       PACKAGE_VERSION is an optional paramater whose default value is 2.13.6"
+
+PACKAGE_VERSION="${1:-$PACKAGE_VERSION}"
+
+yum update -y
+yum install -y git curl
+
+#install java
+yum install -y java java-devel
+which java
+ls /usr/lib/jvm/
+export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.312.b07-1.el8_4.ppc64le
+
+#update path
+export PATH=$PATH:$JAVA_HOME/bin
+
+#install sbt
+rm -f /etc/yum.repos.d/bintray-rpm.repo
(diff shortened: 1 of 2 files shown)


s/scala-library/scala-library_ubi8.sh
```diff
@@ -0,0 +1,58 @@
...
+#!/bin/bash
+
+PACKAGE_VERSION=2.13.6
+
+echo "Usage: $0 [-v <PACKAGE_VERSION>]"
+echo "       PACKAGE_VERSION is an optional paramater whose default value is 2.13.6"
+
+PACKAGE_VERSION="${1:-$PACKAGE_VERSION}"
+
+yum update -y
```
amitsadaphule, line 26: Is this `yum update` really needed? If not, can you please remove it?

s/scala-library/scala-library_ubi8.sh
```diff
@@ -0,0 +1,58 @@
...
+yum install -y git curl
+
+#install java
+yum install -y java java-devel
+which java
+ls /usr/lib/jvm/
+export JAVA_HOME=/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.312.b07-1.el8_4.ppc64le
```
amitsadaphule, line 33: It's better to install specific version of openjdk/java. In this particular case `java-1.8.0-openjdk-devel`.

Also, assigning absolute path to `JAVA_HOME` may create issues later due to availability of a higher sub-version, causing the path in `JAVA_HOME` non-existent.

You can use [this](https://github.com/ppc64le/build-scripts/blob/master/p/pip-ray/pip-ray_0.7.7_rhel7.6.sh#L26-L28) as reference.

s/scala-library/scala-library_ubi8.sh
```diff
@@ -0,0 +1,58 @@
...
+cd scala
+git checkout v$PACKAGE_VERSION
+sbt compile
+
+
+#sample project using scala to test
+mkdir testpro
+cd testpro
+sbt new scala/hello-world.g8
+#input 'hello world' when prompted
```
amitsadaphule, line 54: Please see if it is possible to bypass this user intervention.

amitsadaphule (changes requested):

amitsadaphule:
Closing this on account of https://github.com/ppc64le/build-scripts/pull/1146

PR #1172, opened 2021-11-16
Added build-scripts for 4 packages.

varsha-aaynure:
Created build-script for these packages:
1. caarlos0-env
2. jaeger-core
3. micrometer-core
4. vavr

### DIFF
--- a/c/caarlos0-env/caarlos0-env_ubi8_3.sh
+++ b/c/caarlos0-env/caarlos0-env_ubi8_3.sh
@@ -1,0 +1,32 @@
+# ----------------------------------------------------------------------------
+#
+# Package       : caarlos0/env [caarlos0-env] 
+# Version       : v6.7.2
+# Tested on     : UBI 8.3
+# Script License: Apache-2.0 License    
+# Maintainer    : Varsha Aaynure <Varsha.Aaynure@ibm.com>
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+#!/bin/bash
+
+yum update -y 
+
+#Install required files
+yum install -y golang
+
+#Cloning Repo
+go get github.com/caarlos0/env/v6
+cd /root/go/pkg/mod/github.com/caarlos0/env/v6\@v6.7.2/
+
+#Build test package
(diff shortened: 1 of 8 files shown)


c/caarlos0-env/caarlos0-env_ubi8_3.sh
```diff
@@ -0,0 +1,32 @@
...
+
+#!/bin/bash
+
+yum update -y 
+
+#Install required files
+yum install -y golang
+
+#Cloning Repo
+go get github.com/caarlos0/env/v6
```
kandarpamalipeddi, line 25: Correct me if I am wrong.
Looks like, This always gets the latest V6 version (in future it would be v6.7.3 or v6.7.4).
It won't be specific to version v6.7.2.

So my suggestion is to go with `git clone` and `git checkout` the required version.
Also take the version as a variable, something like : `VERSION=${1:-v6.7.2}`

j/jaeger-core/jaeger-core_ubi8_3.sh
```diff
@@ -0,0 +1,38 @@
...
+
+yum update -y 
+
+#Install required files
+yum install -y git 
+
+#Cloning Repo
+git clone $PACKAGE_URL
+cd jaeger-client-java/jaeger-core/
+git checkout $PACKAGE_VERSION
```
kandarpamalipeddi, line 32: Can we checkout a specific tag/branch/commit-id instead of master ?

m/micrometer-core/micrometer-core_ubi8_3.sh
```diff
@@ -0,0 +1,37 @@
...
+
+yum update -y 
+
+#Install required files
+yum install -y git 
+
+#Cloning Repo
+git clone $PACKAGE_URL
+cd micrometer/
+git checkout
```
kandarpamalipeddi, line 31: Can we checkout a specific tag/branch/commit-id instead of master ?

v/vavr/vavr_ubi8_3.sh
```diff
@@ -0,0 +1,38 @@
...
+
+yum update -y 
+
+#Install required files
+yum install -y git java-1.8.0-openjdk-devel
+
+#Cloning Repo
+git clone $PACKAGE_URL
+cd vavr/
+git checkout $PACKAGE_VERSION
```
kandarpamalipeddi, line 32: Can we checkout a specific tag/branch/commit-id instead of master ?

kandarpamalipeddi (changes requested):

PR #1181, opened 2021-11-18
use correct git url for sentry sdk build script

Sapana-Khemkar:


s/sentry-python/sentry-python_ubi_8.4.sh
```diff
@@ -28,11 +28,11 @@ fi
 
 # Dependency installation
 dnf update
```
amitsadaphule, line 30: Is this `dnf update` really necessary? If yes, please add `-y` to the command so that it does not get blocked for user input, otherwise please remove it.

amitsadaphule (changes requested):
Except for the `dnf update` query, rest looks OK.

s/sentry-python/sentry-python_ubi_8.4.sh
Sapana-Khemkar: dnf update not necessary. Hence removed.

amitsadaphule (approved):

PR #1198, opened 2021-11-24
Added build script for opm

hari291:
opm.sh


snehlatamohite (changes requested):
I see 24 files changed. Do u want to commit all of them or script for opm binary?

seth-priya:
@hari291 any update?

hari291:
@seth-priya @snehlatamohite Raised PR only for opm script

PR #1200, opened 2021-11-24
Adding node.js (v12.16.0) build script

Manik-ibm:


seth-priya:
why do we need this? the ppc64le binaries are already officially available at https://nodejs.org/download/release/v12.16.0/ (cc @snehlatamohite ) ...

PR #1208, opened 2021-11-26
build script for jamm version 0.3.0 (with jam patch)  and master

Sapana-Khemkar:


j/jamm/jamm-master-ubi8.4.sh
```diff
@@ -0,0 +1,37 @@
+# -----------------------------------------------------------------------------
+#
+# Package	: jamm
+# Version	: master
```
kandarpamalipeddi, line 4: Any specific reason for building master ?
Because, master branch will keep on changing and we can't say master branch build is stable.

Sapana-Khemkar: Requested version was 0.3.0 which is from 2014. This version has issue and need fix in maven url in build.xml. Hence as suggested by Priya created two scripts one for master without any patch and another for requested version with patch.

kandarpamalipeddi: Okay. Thanks for the details.

In that case, can you check out the current commit id on master branch (instead of using master, use a latest commit id) as version ? That makes our build script stable, even if anything added/failed on master in future, it won't effect our build script.

j/jamm/jamm-master-ubi8.4.sh
```diff
@@ -0,0 +1,37 @@
...
+# ----------------------------------------------------------------------------
+set -e
+
+PACKAGE_NAME="jamm"
+PACKAGE_VERSION="v0.3.0"
+PACKAGE_URL="https://github.com/jbellis/jamm.git"
+APACHE_ANT_VERSION="1.10.12"
+
+#install dependencies
+yum install  -y java-1.8.0-openjdk java-1.8.0-openjdk-devel java-1.8.0-openjdk-headless gcc-c++ wget unzip
```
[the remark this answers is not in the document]
Sapana-Khemkar: agree gcc-c++, wget and unzip not required for this script. Changes done and committed

j/jamm/jamm-v0.3.0-ubi8.4.sh
```diff
@@ -0,0 +1,54 @@
...
+rm -f apache-ant-$APACHE_ANT_VERSION-bin.zip 
+
+ANT_HOME=/home/tester/apache-ant-$APACHE_ANT_VERSION
+PATH=$PATH:$ANT_HOME/bin
+
+#clone jamm
+git clone $PACKAGE_URL 
+cd $PACKAGE_NAME
+git checkout tags/$PACKAGE_VERSION
+git apply $PATCH_PATH
```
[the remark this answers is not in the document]
Sapana-Khemkar: I think no as I am not raising PR for package but @seth-priya can confirm.

j/jamm/jamm-master-ubi8.4.sh
Sapana-Khemkar: I have tried with latest stable version 0.3.3. But this version also has same issue. Only master code is building without patch.

Sapana-Khemkar: changes committed

Sapana-Khemkar:
As patch needs OSSC approval, removed patch and script for v0.3.0

seth-priya (approved):

PR #1219, opened 2021-11-30
Added build script for etcd

atharv-phadnis:
Package Name: **etcd**
Package Version: **v3.5.1**
Package URL: [etcd](https://github.com/etcd-io/etcd)

Tested on: **UBI 8.5**
Build Errors: **None**
Test Errors: **None**

Comments: Version Upgraded **v3.3.13 -> v3.5.1**

### DIFF
--- a/e/etcd/etcd_ubi_8.5.sh
+++ b/e/etcd/etcd_ubi_8.5.sh
@@ -1,0 +1,56 @@
+# -----------------------------------------------------------------------------
+#
+# Package	: etcd
+# Version	: v3.5.1
+# Source repo	: https://github.com/etcd-io/etcd
+# Tested on	: UBI 8.5
+# Script License: Apache License, Version 2 or later
+# Maintainer	: Atharv Phadnis <Atharv.Phadnis@ibm.com>
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+PACKAGE_NAME=etcd
+PACKAGE_VERSION=v3.5.1
+PACKAGE_URL=https://github.com/etcd-io/etcd
+
+yum install -y git golang make
+
+HOME_DIR=`pwd`
+
+OS_NAME=$(cat /etc/os-release | grep ^PRETTY_NAME | cut -d= -f2)
+
+if ! git clone $PACKAGE_URL $PACKAGE_NAME; then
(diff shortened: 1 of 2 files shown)


e/etcd/etcd_ubi_8.5.sh
```diff
@@ -0,0 +1,56 @@
...
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+PACKAGE_NAME=etcd
+PACKAGE_VERSION=v3.5.1
```
kandarpamalipeddi, line 19: Make PACKAGE_VERSION supplied from command line arguments.
Something like :
`PACKAGE_VERSION=${1:-v3.5.1}`

e/etcd/etcd_ubi_8.5.sh
```diff
@@ -0,0 +1,56 @@
...
+PACKAGE_URL=https://github.com/etcd-io/etcd
+
+yum install -y git golang make
+
+HOME_DIR=`pwd`
+
+OS_NAME=$(cat /etc/os-release | grep ^PRETTY_NAME | cut -d= -f2)
+
+if ! git clone $PACKAGE_URL $PACKAGE_NAME; then
```
kandarpamalipeddi, line 28: In my opinion, while dealing with go packages, try to clone and build under GOPATH.
```
export GOPATH=$HOME/go
mkdir -p $GOPATH/src/github.com/etcd-io/
cd $GOPATH/src/github.com/etcd-io/
git clone $PACKAGE_URL $PACKAGE_NAME
```
Please go through the GO documentation for `GOPATH` and `GOROOT`

kandarpamalipeddi (changes requested):

e/etcd/etcd_ubi_8.5.sh
atharv-phadnis: @kandarpamalipeddi Actually in case of this package, it generates a directly usable binary, instead of being _installed as a go package_; so I built it by directly cloning (to run from /etcd/bin). Should I still do that in `GOPATH` instead?

kandarpamalipeddi: Okay.
I feel it is better to keep it in GOPATH. As it is a binary, its up to you for making changes. :)

atharv-phadnis: Made changes according to your suggestions, please review

kandarpamalipeddi (approved):

PR #1220, opened 2021-11-30
Build scripts for NPM packages

gitssinghal:


kandarpamalipeddi (changes requested):
@gitssinghal 
Please confirm, do we need `wget gcc-c++ make python2` for building npm packages ?

Also, check for the checkout versions 
Package [sinon-chai](https://github.com/domenic/sinon-chai), does not have `v3.3.0`, it have `3.3.0`. 
Git checkout will fail, can you please check for the others as well ?

PR #1222, opened 2021-12-01
Build scripts for soap-builder

sapnashukla28:
Package: Soap-builder
url: https://github.com/reficio/soap-ws

Changes made:

- Kept version as master as the given repo only has a master branch, there are no other tags or branches.
- As soap-builder build is interdependent on other packages, first built the main repo(soap-ws) itself skipping the test to increase the speed. And also added -T 1C to include 1thread per core to increase build speed.

Thanks,
Sapna.

### DIFF
--- a/s/soap-builder/soap-builder_ubi8.4.sh
+++ b/s/soap-builder/soap-builder_ubi8.4.sh
@@ -1,0 +1,48 @@
+# ----------------------------------------------------------------------------
+#
+# Package       : soap-builder
+# Version       : 1.0.0-SNAPSHOT
+# Source repo   : https://github.com/reficio/soap-ws
+# Tested on     : ubi: 8.4
+# Script License: Apache License 2.0
+# Maintainer's  : Sapna Shukla<Sapna.Shukla@ibm.com>
+#
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+#!/bin/bash
+
+#Variables.
+BASE_PACKAGE_NAME=soap-ws
+PACKAGE_NAME=soap-builder
+PACKAGE_URL=https://github.com/reficio/soap-ws
+PACKAGE_VERSION=master
+
+# Installation of required sotwares. 
+yum install git maven -y 
+
+# Cloning the repository from remote to local. 
+git clone $PACKAGE_URL
+cd $BASE_PACKAGE_NAME
+git checkout $PACKAGE_VERSION
+
+# Build and test.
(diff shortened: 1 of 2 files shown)


s/soap-builder/soap-builder_ubi8.4.sh
```diff
@@ -0,0 +1,49 @@
...
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+#!/bin/bash
+
+#Variables.
+BASE_PACKAGE_NAME=soap-ws
+PACKAGE_NAME=soap-builder
+PACKAGE_URL=https://github.com/reficio/soap-ws
+PACKAGE_VERSION=master
```
amitsadaphule, line 24: The versions in header and here (`master`) do not match. In case of top-of-tree i.e. master, please mention a specific commit # against which the script has been verified.

s/soap-builder/soap-builder_ubi8.4.sh
```diff
@@ -0,0 +1,49 @@
...
+cd $BASE_PACKAGE_NAME
+git checkout $PACKAGE_VERSION
+
+# Build and test.
+if ! mvn -T 1C clean install -Dlicense.skip=true -Dadditionalparam=-Xdoclint:none -DskipTests; then
+	echo "------------------$BASE_PACKAGE_NAME:INSTALL_FAILED-------------------------------------"
+	exit 1
+fi
+cd $PACKAGE_NAME
+if ! mvn clean install -Dlicense.skip=true; then
```
amitsadaphule, line 40: Are we trying to build the sub-module as a fallback here in case the full base package build fails?

amitsadaphule (changes requested):

s/soap-builder/soap-builder_ubi8.4.sh
sapnashukla28: Yes, changed the version from master to the commit ID from the top of the tree.

s/soap-builder/soap-builder_ubi8.4.sh
sapnashukla28: Yes, just to make sure that the sub-modules compile correctly before running the tests.

s/soap-builder/soap-builder_ubi8.4.sh
```diff
@@ -1,7 +1,7 @@
 # ----------------------------------------------------------------------------
 #
 # Package       : soap-builder
-# Version       : 1.0.0-SNAPSHOT
+# Version       : 93e5a7956ed68b6125ab756e9c6cddea266a9dc6 (commit ID)
```
amitsadaphule, line 4: We specify commit SHAs in the following format in general:
```
commit # 93e5a7956ed68b6125ab756e9c6cddea266a9dc6
```
Please use this format in the future.

amitsadaphule (approved):
LGTM

PR #1227, opened 2021-12-02
Upath ubi 8.4

vikasgupta8:
Have set default version to 1.2.0 and can supply other versions through command line.
Tested with v1.2.0 and v2.0.1


u/upath/upath_ubi_8.4.sh
```diff
@@ -0,0 +1,64 @@
+# -----------------------------------------------------------------------------
+#
+# Package       : upath
+# Version       : v1.2.0, v2.0.1
+# Source repo   : https://github.com/anodynos/upath
+# Tested on     : RHEL ubi 8.4
```
amitsadaphule, line 6: Does this mean that the script has been validated on RHEL 8.4 as well as UBI 8.4? The scripts which work on UBI 8.4 almost certainly work on RHEL 8.4, but the reverse is not always true.

u/upath/upath_ubi_8.4.sh
```diff
@@ -0,0 +1,64 @@
...
+        exit 0
+fi
+
+cd $HOME_DIR/$PACKAGE_NAME/
+
+git checkout $PACKAGE_VERSION
+PACKAGE_VERSION=$(jq -r ".version" package.json)
+
+npm install -g grunt && npm install -g grunt-cli
+cd $HOME_DIR/$PACKAGE_NAME/
```
amitsadaphule, line 44: We're already inside `$HOME_DIR/$PACKAGE_NAME/` as per line 38.

u/upath/upath_ubi_8.4.sh
```diff
@@ -0,0 +1,64 @@
...
+        echo "------------------$PACKAGE_NAME:clone_fails---------------------------------------"
+        echo "$PACKAGE_URL $PACKAGE_NAME" > /home/tester/output/clone_fails
+        echo "$PACKAGE_NAME  |  $PACKAGE_URL |  $PACKAGE_VERSION | $OS_NAME | GitHub | Fail |  Clone_Fails" > /home/tester/output/version_tracker
+        exit 0
+fi
+
+cd $HOME_DIR/$PACKAGE_NAME/
+
+git checkout $PACKAGE_VERSION
+PACKAGE_VERSION=$(jq -r ".version" package.json)
```
amitsadaphule, line 41: Why are we setting `PACKAGE_VERSION` again here? Does the tag name does not match with the version from `package.json`?

amitsadaphule (changes requested):

u/upath/upath_ubi_8.4.sh
vikasgupta8: It was tested on ubi 8.4. I have updated the script.

u/upath/upath_ubi_8.4.sh
vikasgupta8: Yes removed the command.

u/upath/upath_ubi_8.4.sh
vikasgupta8: I have removed the re-setting of PACKAGE_VERSION

amitsadaphule (changes requested):
Please resolve the conflicts on the PR.

vikasgupta8:
Resolved conflicts.

amitsadaphule (approved):
LGTM

PR #1242, opened 2021-12-08
Added scripts for rest-assured packages

Balavva-Mirji:
Added scripts for rest-assured packages.


snehlatamohite (approved):
@Balavva-Mirji  - I am not sure why two scripts with same name REST-assured-common_ubi_8.3.sh  are presnt in PR?

PR #1275, opened 2021-12-14
Build script for sha, postcss-selector-parser, is-regex, is-boolean-object, i18next, has-symbols, flux, flatpickr"

gitssinghal:
sha, postcss-selector-parser, is-regex, is-boolean-object, i18next, has-symbols, flux, flatpickr

### DIFF
--- a/h/has-symbols/has-symbols_ubi8.sh
+++ b/h/has-symbols/has-symbols_ubi8.sh
@@ -1,0 +1,41 @@
+# ----------------------------------------------------------------------------
+#
+# Package               : has-symbols
+# Version               : 1.0.2
+# Source repo           : https://github.com/facebook/has-symbols
+# Tested on             : UBI 8
+# Script License        : Apache License, Version 2 or later
+# Maintainer            : Swati Singhal <swati.singhal@ibm.com>
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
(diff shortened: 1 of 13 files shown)


h/has-symbols/has-symbols_ubi8.sh
```diff
@@ -0,0 +1,41 @@
...
+
+NODE_VERSION=v12.22.4
+#installing nvm
+curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+source ~/.bashrc
+nvm install $NODE_VERSION
+
+# clone, build and test specified version
+
+git clone https://github.com/facebook/has-symbols
```
kandarpamalipeddi, line 37: @gitssinghal 
This link even not accessible.

image

s/sha/sha.js
```diff
@@ -0,0 +1,41 @@
+# ----------------------------------------------------------------------------
```
kandarpamalipeddi, line 1: Please rename this file to `sha_v3.0.0_ubi8.sh`

kandarpamalipeddi (commented):
A generic comment.

- We don't need `gcc-c++, wget make and python`  dependencies for any of these packages.
- I see checkout tags for some of the packages are not working. Eg. facebook/flux, available version is `4.0.1`, but in script checkout version mentioned is `v4.0.1`. In this case checkout fails and master will get build.
- No `npm test` added.

**These comments applicable for other PRs as well.**

PR #1287, opened 2021-12-20
script for package classmate

vikasgupta8:
Created build script for package classmate.

### DIFF
--- a/c/classmate/classmate_8.4.sh
+++ b/c/classmate/classmate_8.4.sh
@@ -1,0 +1,49 @@
+#----------------------------------------------------------------------------
+#
+# Package		: classmate
+# Version		: 1.3.4, 1.5.1
+# Source repo		: https://github.com/FasterXML/java-classmate
+# Tested on		: RHEL 8.4
+# Script License	: Apache License Version 2.0
+# Maintainer		: Vikas Gupta <vikas.gupta8@ibm.com>
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+#!/bin/bash
+
+mkdir -p /logs
+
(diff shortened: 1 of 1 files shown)


c/classmate/classmate_8.4.sh
```diff
@@ -0,0 +1,48 @@
...
+LOGS_DIRECTORY=/logs
+LOCAL_DIRECTORY=/home/tester
+REPOSITORY="https://github.com/FasterXML/java-classmate.git"
+
+
+# ------- Clone and build source -------
+
+export PATH=/usr/lib/jvm/java-1.8.0-openjdk-1.8.0.312.b07-2.el8_5.ppc64le/bin/:$PATH
+java -version
+yum install maven
```
kandarpamalipeddi, line 35: Looks like, `yum install maven` should be before line 33 (`export ...`), otherwise, `java -version` will fail to find the java.
Also make it as `yum install -y maven`, so that script will not wait for user input.

kandarpamalipeddi (changes requested):

c/classmate/classmate_8.4.sh
```diff
@@ -0,0 +1,48 @@
+#----------------------------------------------------------------------------
+#
+# Package		: classmate
+# Version		: 1.3.4, 1.5.1
+# Source repo		: https://github.com/FasterXML/java-classmate
+# Tested on		: RHEL 8.4
+# Script License	: Apache License Version 2.0
+# Maintainer		: Vikas Gupta <vikas.gupta8@ibm.com>
```
kandarpamalipeddi, line 8: Include `# Language : Java` in header.

kandarpamalipeddi (changes requested):

c/classmate/classmate_8.4.sh
vikasgupta8: Added Language : Java

c/classmate/classmate_8.4.sh
vikasgupta8: Done.

kandarpamalipeddi (approved):

PR #1290, opened 2021-12-20
Build script to build and validate package etcd

vikasgupta8:
Build script to build and validate package etcd.
Requested version is initial version v0.5.0 and was build in 2014 around 7-8 years back.
We have written the script for latest stable version v3.5.1.

### DIFF
--- a/e/etcd/etcd_ubi8.4.sh
+++ b/e/etcd/etcd_ubi8.4.sh
@@ -1,7 +1,7 @@
 # -----------------------------------------------------------------------------
 #
 # Package       : github.com/etcd-io/etcd and go.etcd.io/etcd
-# Version       : 0, v0.1.0 and v0.5.0
+# Version       : v3.5.1
 # Source repo   : https://github.com/etcd-io/etcd
 # Tested on     : RHEL 8.4
 # Script License: Apache License, Version 2 or later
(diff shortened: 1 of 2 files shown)


e/etcd/etcd_ubi8.4.sh
```diff
@@ -0,0 +1,63 @@
+# -----------------------------------------------------------------------------
+#
+# Package       : github.com/etcd-io/etcd and go.etcd.io/etcd
+# Version       : v3.5.1
+# Source repo   : https://github.com/etcd-io/etcd
+# Tested on     : RHEL 8.4
```
seth-priya, line 6: If we have tested this in a UBI container, let's mention the distro as UBI 8.4. If it has been tested on RHEL as well, let's specific both the distros with versions as a comma separated list.

The script name should be "etcd_ubi_8.4.sh" or "etcd_rhel_8.4.sh" accordingly

seth-priya (changes requested):

e/etcd/etcd_ubi8.4.sh
vikasgupta8: updated distro version to ubi only.

seth-priya:
The script for this for same version is already now present at https://github.com/ppc64le/build-scripts/blob/master/e/etcd/etcd_ubi_8.5.sh , hence closing this PR

PR #1321, opened 2021-12-28
Added build script for thrift

atharv-phadnis:
Package Name: **thrift**
Package Version: **v0.13.0**
Package URL: [thrift](https://github.com/apache/thrift)

Tested on: **UBI 8.5**
Build Errors: **None**
Test Errors: **None**

### DIFF
--- a/t/thrift/thrift_ubi_8.5.sh
+++ b/t/thrift/thrift_ubi_8.5.sh
@@ -1,0 +1,94 @@
+# -----------------------------------------------------------------------------
+#
+# Package	: thrift
+# Version	: v0.13.0
+# Source repo	: https://github.com/apache/thrift
+# Tested on	: UBI 8.5
+# Script License: Apache License, Version 2 or later
+# Maintainer	: Atharv Phadnis <Atharv.Phadnis@ibm.com>
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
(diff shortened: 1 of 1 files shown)


t/thrift/thrift_ubi_8.5.sh
```diff
@@ -0,0 +1,94 @@
...
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+PACKAGE_NAME=thrift
+PACKAGE_VERSION=${1:-v0.13.0}
+PACKAGE_URL=https://github.com/apache/thrift
+
+# Include CentOS Repos
+dnf -y install \
```
seth-priya, line 23: @atharv-phadnis is there a way to avoid using the centos repositories? which package (s) are being installed from there? We are not sure if every one would be open to enabling this on RHEL/UBI, also can you check if these are available on RHEL?

atharv-phadnis: @seth-priya It was required for installation of group "Development Tools", which wasn't available in the UBI repos. Main requirement from the groups is the **bison** package, I've tried building it from source, it gets build and installed successfully, but results in error during the building of **thrift** itself.

atharv-phadnis: Updated the script as per your suggestions, please review.

seth-priya (approved):

PR #1323, opened 2021-12-29
Added 4 node scripts which are built on ubi 8.3

Balavva-Mirji:
Added below 4 node scripts
http2-client
istanbul-lib-coverage
istanbul-lib-instrument
jsonpath

### DIFF
--- a/h/http2-client/LICENSE
+++ b/h/http2-client/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
+   1. Definitions.
+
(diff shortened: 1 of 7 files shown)


seth-priya:
@Balavva-Mirji , couple of things,
- please use node12, instead of node10 (which is quite old) in all the scripts - you can use "dnf module enable nodejs:12" followed by "dnf install nodejs"
- use "set -e" to detect failures

Balavva-Mirji:
@seth-priya - for the package jsonpath, gulp 3 is not compatible with either node 12 or node 14. So can we go with node 10 for this package?

seth-priya:
> @seth-priya - for the package jsonpath, gulp 3 is not compatible with either node 12 or node 14. So can we go with node 10 for this package?

yes, but in such cases explicitly put in a comment in the script mentioning the same

seth-priya (approved):

PR #22, opened 2018-11-28
Docker files for TensorFlow Serving

wdirons:
Based on the docker file from:
https://github.com/tensorflow/serving/tree/master/tensorflow_serving/tools/docker

Built bazel from source instead of installing x86 version
Removed TENSORRT as it is not supported on POWER yet
Updated to CUDA 10 and Ubuntu 18.04
Fixed paths for ppc64le
Installed TensorFlow 1.12.0 from OSU release builds as it
can't be installed from pip

Will submit a PR to tensorflow/serving when they move up to CUDA 10.

PR #1343, opened 2022-01-03
Gojsonschema: buildscript

rajusah15:
## Checklist

- [x] Did you have **Legal approvals** for patch files ?

### DIFF
--- a/a/airline/airline_rhel_8.3.sh
+++ b/a/airline/airline_rhel_8.3.sh
@@ -1,0 +1,31 @@
+# -----------------------------------------------------------------------------
+#
+# Package       : airline
+# Version       : 0.6
+# Source repo   : https://github.com/airlift/airline.git
+# Tested on     : UBI 8
+# Script License: Apache License, Version 2 or later
+# Maintainer    : Raju.Sah@ibm.com
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+yum update -y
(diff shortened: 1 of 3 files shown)


seth-priya:
@rajusah15 this one (and some of your other PRs as well) seem to have the build script for not just gsonschema but also airline, was that intentional?

Couple of additional, generic comments that probably apply to the other PRs that you have raised as well:
1- The scripts don't seem to adhere to the latest templates that we have here https://github.com/ppc64le/build-scripts/tree/master/templates, in particular 

a) the Language and Travis-Check fields are missing from the header (see line #7 and #8 https://github.com/ppc64le/build-scripts/blob/66d9be10cc51eecd1859e81cc3f27d17cd4fca25/templates/build_script_go.sh#L7 )

b) the "PACKAGE_NAME" / "PACKAGE_VERSION" and "PACKAGE_URL" variables are all missing (Refer to lines #20 - #22 here https://github.com/ppc64le/build-scripts/blob/66d9be10cc51eecd1859e81cc3f27d17cd4fca25/templates/build_script_go.sh#L20)

c) the build steps don't seem to be consistent / in-sync with the template, we can simplify the template as needed but should try to re-use the steps that can be from there - see some of the recent build scripts for go packages that have been added as an example.

2) The script name says "RHEL 8.3" but the header says UBI 8, please be consistent in specifying the distro name and version or if tested on both, specify both in the header like "RHEL 8.3, UBI 8.3"

3) Please use "set -e" in all the scripts to ensure any intermediate failures are detected.

4) The checkboxes in the PR template should be checked as applicable.

Please implement and address the above for all the PRs before we get deeper into reviewing any of them , cc @kandarpamalipeddi

rajusah15:
@seth-priya @kandarpamalipeddi 
 > For airline package i have raised a separate PR so it can be ignored from this particular PR of gojsonschema.
> I have followed the new template for script.
> I have run the script using 'set -e' option.
> since contribute.md file is not there so followed the build steps provided in the repo.
> Added three required variable which has mentioned in the review.

PR #1354, opened 2022-01-03
Added build script for Elastic

atharv-phadnis:
## Checklist

Package Name: **elastic**
Package Version: **v7.0.22**
Package URL: [elastic](https://github.com/olivere/elastic/)

Tested on: **UBI 8.5**
Build Errors: **None**
Test Errors: **None**

### DIFF
--- a/e/elastic/elastic_ubi_8.5.sh
+++ b/e/elastic/elastic_ubi_8.5.sh
@@ -4,6 +4,8 @@
 # Version	: v7.0.22
 # Source repo	: https://github.com/olivere/elastic
 # Tested on	: UBI 8.5
+# Language      : GO
+# Travis-Check  : False
 # Script License: Apache License, Version 2 or later
 # Maintainer	: Atharv Phadnis <Atharv.Phadnis@ibm.com>
 #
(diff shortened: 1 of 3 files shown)


e/elastic/elastic_ubi_8.5.sh
```diff
@@ -0,0 +1,99 @@
+# -----------------------------------------------------------------------------
+#
+# Package	: github.com/olivere/elastic/v7
+# Version	: v7.0.22
+# Source repo	: https://github.com/olivere/elastic
+# Tested on	: UBI 8.5
+# Script License: Apache License, Version 2 or later
+# Maintainer	: Atharv Phadnis <Atharv.Phadnis@ibm.com>
+#
```
seth-priya, line 9: Looks like Travis-Check should be mentioned and False for this one? The "Language" field is also missing

atharv-phadnis: Travis-Check is confusing in this case. I've tested the script on a container, but with the docker socket exposed to it. Had discussed with Ryan as well, even he was unsure about these cases. Maybe you could confirm if I should set it to True or False considering the scenario.

Will add the Language field along with Travis-Check.

Please give your suggestion...

seth-priya: I don't think Travis-Check will work for this, let's keep it as false @atharv-phadnis

atharv-phadnis: Updated, please review

PR #1368, opened 2022-01-04
Gojsonschemas: buildscript

rajusah15:
## Checklist


g/gojsonschema/gojsonschema_rhel_8.3.sh
```diff
@@ -0,0 +1,39 @@
...
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ------------------------------------------------------------------------------
+set -e
+PACKAGE_NAME=gojsonschema
+PACKAGE_VERSION={1:-v1.2.0}
+PACKAGE_URL=https://github.com/xeipuuv/gojsonschema.git
+
+yum update -y && yum install -y git golang
```
seth-priya, line 24: I don't think we need a "yum update here", any specific reason for including it?

g/gojsonschema/gojsonschema_rhel_8.3.sh
```diff
@@ -0,0 +1,39 @@
...
+
+#Add dependency
+go get github.com/xeipuuv/gojsonschema
```
seth-priya, line 27: do we need to individually install each dependency? can we do something like a "go get -d -t $PACKAGE_NAME@$PACKAGE_VERSION" which would eliminate the need for a git clone as well?

g/gojsonschema/gojsonschema_rhel_8.3.sh
```diff
@@ -0,0 +1,39 @@
+# -----------------------------------------------------------------------------
+#
+# Package       : gojsonschema
+# Version       : v1.2.0
+# Source repo   : https://github.com/xeipuuv/gojsonschema.git
+# Tested on     : RHEL 8.3
```
seth-priya, line 6: have we tested this on a UBI based container environment as well?

g/gojsonschema/gojsonschema_rhel_8.3.sh
rajusah15: 'yum update' can be removed , it will not affect the script.

g/gojsonschema/gojsonschema_rhel_8.3.sh
rajusah15: I have tested on ubi 8.3 so i will change 'Tested on' and file name wilt 'rhel 8.3'

seth-priya:
@rajusah15 no need to specify the version of the package in the script name the build instructions are generic and expected to work for higher versions as well, rest looks good to me

g/gojsonschema/gojsonschema_ubi_8.3.sh
```diff
@@ -0,0 +1,34 @@
...
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ------------------------------------------------------------------------------
+set -e
+PACKAGE_NAME=github.com/xeipuuv/gojsonschema
+PACKAGE_VERSION=${1:-v1.2.0}
+PACKAGE_URL=https://github.com/xeipuuv/gojsonschema.git
```
seth-priya, line 22: oh and missed this earlier - trivial - but probably the package url is no longer needed as we aren't doing git clone?

seth-priya (approved):

PR #1369, opened 2022-01-04
Added ppc64le build script for attrs package

kumarvikas2020:
## Checklist

Package Name: gobuffalo/attrs
Package Version: v0.0.0-20190224210810-a9411de4debd
Package URL: [attrs](https://github.com/gobuffalo/attrs.git)

Tested on: UBI 8.4

### DIFF
--- a/a/attrs/attrs_ubi_8.4.sh
+++ b/a/attrs/attrs_ubi_8.4.sh
@@ -5,6 +5,7 @@
 # Source repo           : https://github.com/gobuffalo/attrs.git
 # Tested on             : UBI 8.4
 # Language              : GO
+# Travis-Check          : True
 # Script License        : Apache License, Version 2 or later
 # Maintainer            : Vikas . <kumar.vikas@in.ibm.com>
 #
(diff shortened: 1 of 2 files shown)


a/attrs/attrs_ubi_8.4.sh
```diff
@@ -0,0 +1,71 @@
...
+  export VERSION=a9411de4debd
+else
+  export VERSION=$1
+fi
+if [ -d "attrs" ] ; then
+  rm -rf attrs
+fi
+
+# Dependency installation
+dnf install -y git gcc
```
seth-priya, line 30: do we need gcc? in general it would be good to install only the minimal set of dependencies needed to build and test the package

kumarvikas2020: gcc is required to build the package:
```
cgo: exec gcc: exec: "gcc": executable file not found in $PATH
FAIL    github.com/gobuffalo/attrs [build failed]
```

seth-priya (approved):

PR #1423, opened 2022-01-11
Added build script for evanphx/json-patch

vathsalashetty25:
## Checklist
Package Name: evanphx/json-patch
Package Version: v0.0.0-20190203023257-5858425f7550
Package URL: https://github.com/evanphx/json-patch

Tested on: UBI 8.5


g/github.com__evanphx__json-patch/github.com__evanphx_json-patch_0.0.0-20190203023257-5858425f7550_rhel_8.4.sh
```diff
@@ -0,0 +1,82 @@
+# ----------------------------------------------------------------------------
+#
+# Package               : Json-patch
+# Version               : v0.0.0-20190203023257-5858425f7550
```
seth-priya, line 4: does this same script work for top of tree as well? If so, we can remove the version info. from the script name

vathsalashetty25: Yes the same script work for top of tree as well,i've removed the version info from the script name.

PR #1428, opened 2022-01-11
Added build script for evanphx/json-patch v4.5.0

vathsalashetty25:
## Checklist
Package Name: evanphx/json-patch
Package Version: v4.5.0
Package URL: https://github.com/evanphx/json-patch

Tested on: UBI 8.5


seth-priya:
@vathsalashetty25 is the script specific to the version tested with or does it work for top of tree? If yes, we can remove the version number from the script name making it more generic

seth-priya:
and I guess only one of https://github.com/ppc64le/build-scripts/pull/1423 or this would be needed if there are no differences?

vathsalashetty25:
I have changed the script name..Yes this script works for top of tree as well...so can i delete one PR?

PR #1561, opened 2022-01-20
Travis-Check flag check added, along with review suggestions

Ryan-Dmello:
## Checklist

Changes consists:
1. Additional check to verify if the package name, package version and package URL in doc string matches with the information in the script.
2. Build script validation will be triggered only if Travis-Check flag is set to True. If set to False the script validation will be skipped and build will return successful.

### DIFF
--- a/script/validate_builds.py
+++ b/script/validate_builds.py
@@ -20,6 +20,8 @@ def trigger_basic_validation_checks(file_name):
         "# Language": "package_type",
         "# Travis-Check": ""
     }
+    should_travis_run_for_script = True
+    package_name = package_version = package_url = None
     matched_keys = []
     # Check if apache license file exists
     file_parts = file_name.split('/')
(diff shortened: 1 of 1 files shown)


script/validate_builds.py
```diff
@@ -48,19 +50,51 @@ def trigger_basic_validation_checks(file_name):
...
+            if package_name and 'PACKAGE_NAME=' in line:
+                key_val = line.split('=')
+                alt_package_name_value = "${1:-%s}"%package_name
+                if str(package_name) != str(key_val[-1].strip()) and str(alt_package_name_value) != str(key_val[-1].strip()):
+                    raise ValueError("PACKAGE_NAME in script, does not match the Package in doc string")
+            elif package_version and 'PACKAGE_VERSION=' in line:
+                key_val = line.split('=')
+                alt_version_value = "${1:-%s}"%package_version
+                if str(package_version) != str(key_val[-1].strip()) and str(alt_version_value) != str(key_val[-1].strip()):
+                    raise ValueError("PACKAGE_VERSION in script, does not match the Version in doc string")
```
kandarpamalipeddi, line 78: There is a chance that, developer can commit a build-script for multiple versions.
In that case,  below is the string:
`# Version : version01, version02`

While setting the package version, it is possible for having a single version as Default version, 
`PACKAGE_VERSION=${1:-version01}`.

Above check will fail at this case.

As we are mandating the package version is configurable, we allow only, `PACKAGE_VERSION=${1:-version01}` format.

so can we have condition something like:
```
if package_version and ('PACKAGE_VERSION=' in line or 'VERSION=' in line):
    pattern = re.compile("(PACKAGE_VERSION|VERSION)=\${\d*:-(?P<version>[^}]*?)}")
    match = pattern.match(line)
    default_version = match.group("version")
    if default_version:
        if default_version not in [x.strip() for x in package_version.split (',')]:
            raise ValueError("PACKAGE_VERSION/VERSION in script, does not match the Version in doc string")
    else:
        raise ValueError("PACKAGE_VERSION/VERSION in script, does not have correct format")
```

CC: @seth-priya , Please correct me if I am wrong.

kandarpamalipeddi (changes requested):
@seth-priya 
Before merging this PR, we need to update the team to follow the pattern,

There should be 3 variable should be defined and should be in-lined with documentation/header section:
`PACKAGE_NAME`
`PACKAGE_VERSION`
`PACKAGE_REPO`

kandarpamalipeddi:
Closing this PR.
We need to discuss more on this before adding more restrictions.

PR #1685, opened 2022-02-01
No validation for scripts under Docekrfile directory, List successful…

kandarpamalipeddi:
…ly validated files.

## Checklist

@seth-priya  @Ryan-Dmello 
Added changes for:

- Not to validated the build-script under Dockerfile directory.
- List out the successfully validated scripts at the end. This is required if no scripts available in the PR (if script extension is other than .sh).

### DIFF
--- a/script/validate_builds.py
+++ b/script/validate_builds.py
@@ -103,7 +103,7 @@ def trigger_build_validation_travis(pr_number):
     for i in response:
         file_name = i.get('filename', "")
         status = i.get('status', "")
-        if file_name.endswith('.sh') and "Dockerfile" not in file_name and status != "removed":
+        if file_name.endswith('.sh') and "dockerfile" not in file_name.lower() and status != "removed":
             # perform basic validation check
             trigger_basic_validation_checks(file_name)
             # Build/test script files
--- a/script/validate_builds.py
+++ b/script/validate_builds.py
@@ -97,16 +97,25 @@ def trigger_build_validation_travis(pr_number):
         GITHUB_BUILD_SCRIPT_BASE_REPO,
         pr_number
     )
+    validated_file_list = []
     response = requests.get(pull_request_file_url).json()
     # Trigger validation for all shell scripts
     for i in response:
         file_name = i.get('filename', "")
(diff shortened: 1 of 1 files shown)


script/validate_builds.py
```diff
@@ -97,16 +97,25 @@ def trigger_build_validation_travis(pr_number):
...
         pr_number
     )
+    validated_file_list = []
     response = requests.get(pull_request_file_url).json()
     # Trigger validation for all shell scripts
     for i in response:
         file_name = i.get('filename', "")
         status = i.get('status', "")
-        if file_name.endswith('.sh') and status != "removed":
+        if file_name.endswith('.sh') and "Dockerfile" not in file_name and status != "removed":
```
Ryan-Dmello, line 106: Do we really need Dockerfile check? can you point to any example where current functionality is trying to run Dockerfile?

As per my understanding, current functionality only triggers shell scripts.

kandarpamalipeddi: As far as I know, when ever we publish Dockerfiles & supporting scripts (something like docker_entrypoint.sh) for a package/component we will add them under `Dockerfile/Dockerfiles` directory. (Eg : [Zookeeper](https://github.com/ppc64le/build-scripts/tree/master/z/zookeeper)).

So basically, I am checking that, if the script name containing "Dockerfile(s)" would be the script added for Docker and no need to validate them.

In case of Zookeeper,  docker-entrypoint.sh is added and [Travis validation](https://app.travis-ci.com/github/ppc64le/build-scripts/builds/245640734) failed (In this case it is failed with basic validation, but I am sure it will fail even if we provide proper header/documentation).

Ryan-Dmello: update the line as below, to ensure case sensitive handling of dockerfile.
`if file_name.endswith('.sh') and "dockerfile" not in file_name.lower() and status != "removed":`

Other changes looks ok

Ryan-Dmello (approved):

kandarpamalipeddi:
Changes made as per comment ...

seth-priya (approved):

PR #1703, opened 2022-02-02
Update and rename node-http-proxy-agent_ubi_8.5.bash to node-http-proxy-agent_ubi_8.5.sh

kumarvikas2020:
## Checklist

Updated file name so that Travis can execute it. Also downgraded to v12.22.9 version of node as latest version is causing issue in Travis job.


kandarpamalipeddi (changes requested):
We already had this script, only change is node version.
Which version of node should be consider ?

kumarvikas2020:
> We already had this script, only change is node version. Which version of node should be consider ?

Any version of node will work. Installing the latest version of node (v17.4.0) was causing issue in Travis CI, so have downgraded to v12.22.9. Also the script passed earlier because the file extension was incorrect.

kandarpamalipeddi (approved):

PR #1823, opened 2022-03-01
Symfony  process ubi 8.4

ApurvaAgrawal3:
## Checklist

### DIFF
--- a/s/symfony__process/symfony__process_rhel_8.3.sh
+++ b/s/symfony__process/symfony__process_rhel_8.3.sh
@@ -6,7 +6,7 @@
 # Source repo	: https://github.com/symfony/process
 # Tested on	: RHEL 8.3, UBI 8.4
 # Language      : PHP
-# Travis-Check  : True
+# Travis-Check  : False
 # Script License: Apache License, Version 2 or later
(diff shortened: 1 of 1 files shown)


seth-priya:
@vedang-wartikar please review

vedang-wartikar:
The package is running successfully when I'm running it locally inside a UBI container. This seems to be a CI job related issue which might be solved after changing some Travis configurations.

Another possible solution would be to set the tty variable to **false**, especially for CI jobs (refer [this issue](https://github.com/deployphp/deployer/issues/1713#issuecomment-425420557))

Going through the symfony source code, I found some variables which could be altered to change the value to false but not sure if this would be the correct approach to follow
```
[root@p006vm88 symfony]# grep -rnw . -e "setTty"
./process/Tests/ProcessTest.php:480:        $process->setTty(true);
./process/Tests/ProcessTest.php:495:        $process->setTty(true);
./process/Tests/ProcessTest.php:510:        $process->setTty(false);
./process/Tests/ProcessTest.php:511:        $process->setTty(true);
./process/Process.php:1065:    public function setTty($tty)
```

ApurvaAgrawal3:
The build script is failing because TTY doesn't supported on travis. On fresh container the build and test are successfully installed but on travis two test case are failing because tty support. I have marked Travis as False in build script. Please review.

kandarpamalipeddi (approved):

PR #1928, opened 2022-03-31
Added build script for netns

anup-kodlekere:
## Checklist

Travis-Check is set to false as package requires additional privileges that can't be run in a container

### DIFF
--- a/n/netns/LICENSE
+++ b/n/netns/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
+   TERMS AND CONDITIONS FOR USE, REPRODUCTION, AND DISTRIBUTION
+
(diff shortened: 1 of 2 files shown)


seth-priya:
@anup-kodlekere can you elaborate on "Travis-Check is set to false as package requires additional privileges that can't be run in a container" , do we know exactly what privileges  are needed?

anup-kodlekere:
> @anup-kodlekere can you elaborate on "Travis-Check is set to false as package requires additional privileges that can't be run in a container" , do we know exactly what privileges are needed?

So I'm not really sure what the required privileges are, but I've tried running in a container, which results in an `Operation not permitted` error, even after running with sudo. 

[Here's](https://github.com/vishvananda/netns) the command that's mentioned on their repo: 
```sudo -E go test github.com/vishvananda/netns```

PR #1950, opened 2022-04-06
Added Dockerfile for Kibana v8.1.0

kumarvikas2020:
## Checklist

### DIFF
--- a/k/kibana/Dockerfiles/8.1.0_ubi_8/Dockerfile
+++ b/k/kibana/Dockerfiles/8.1.0_ubi_8/Dockerfile
@@ -1,0 +1,182 @@
+# NOTE: This is based on the original work at https://github.com/elastic/dockerfiles/
+
+################################################################################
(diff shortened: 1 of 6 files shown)


kumarvikas2020:
@seth-priya 
1 - The build instructions have been added.

2 - This a multistage Dockerfile.
In the builder stage we are extracting the  build/target artifacts to /usr/share/kibana folder:
```
RUN tar -xzf /kibana/target/kibana-8.1.0-SNAPSHOT-linux-ppc64.tar.gz -C /usr/share/kibana --strip-components 1
```
And in the final stage we are copying only the build/target artifacts from the builder stage:
```
COPY --from=builder --chown=1000:0 /usr/share/kibana /usr/share/kibana
```
So the final image is already not having any source code or any other build dependencies.

The size of final image is ~950 MB

seth-priya (approved):

PR #1972, opened 2022-04-08
graylogdoc:added

ankitpar:
## Checklist


seth-priya:
So first thing @ankitpar , looking at these lines in the dockerfile

COPY docker-entrypoint.sh /
COPY health_check.sh /

I don't think the container can be built without these files? If that's the case they need to be included in the PR as well

@prashantkhoje

PR #2070, opened 2022-05-16
Added clickhouseDB build script

pranavpandit1:
## Checklist

### DIFF
--- a/c/clickhouse/clickhouse_ubi8.5.sh
+++ b/c/clickhouse/clickhouse_ubi8.5.sh
@@ -1,0 +1,60 @@
+#!/bin/bash -e
+# ----------------------------------------------------------------------------
+#
+# Package       : ClickHouse
+# Version       : Latest(Top of Tree)
+# Source repo   : https://github.com/ClickHouse/ClickHouse
+# Tested on     : UBI 8.5
+# Language      : C,C++
+# Travis-Check  : True
+# Script License: Apache License, Version 2 or later
+# Maintainer    : Pranav Pandit <pranav.pandit1@ibm.com>
+#
+# Disclaimer: This script has been tested in non-root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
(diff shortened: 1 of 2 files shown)


c/clickhouse/clickhouse_ubi8.5.sh
```diff
@@ -0,0 +1,60 @@
...
+  echo "Build already present "
+else
+  mkdir build
+fi
+cd build
+
+export CC=clang CXX=clang++
+cmake ..
+
+ninja
```
seth-priya, line 60: @pranavpandit1 will this take care of running the unit tests as well?

c/clickhouse/clickhouse_ubi8.5.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash -e
+# ----------------------------------------------------------------------------
+#
+# Package       : ClickHouse
+# Version       : Latest(Top of Tree)
+# Source repo   : https://github.com/ClickHouse/ClickHouse
+# Tested on     : UBI 8.5
+# Language      : C,C++
+# Travis-Check  : True
```
seth-priya, line 9: given that this can not currently run on travis due to the build time limit being exceeded. let's set this to false. Let's add comments to explain the same here

seth-priya, line 9: given that this can not currently run on travis due to the build time limit being exceeded. let's set this to false. Let's add comments to explain the same here

c/clickhouse/clickhouse_ubi8.5.sh
pranavpandit1: No, but power builds are directly available from the community and this build script is just an additional way to build a binary, some exceptions on Unit tests are observed but have not been analyzed in-depth for parity with Intel.

c/clickhouse/clickhouse_ubi8.5.sh
pranavpandit1: Done

pranavpandit1: Done

pranavpandit1: Done

PR #2142, opened 2022-05-30
AddedMInkScriptUBI8.5

ankitpar:
## Checklist

### DIFF
--- a/m/mink-selenium2-driver/minkseleniumdriver_ubi8.5.sh
+++ b/m/mink-selenium2-driver/minkseleniumdriver_ubi8.5.sh
@@ -1,0 +1,71 @@
+#!/bin/bash -e
+
+# ----------------------------------------------------------------------------
+# Package          : php-jwt
+# Version          : v6.2.0
+# Source repo      : https://github.com/firebase/php-jwt
+# Tested on        : UBI 8.5
+# Language         : PHP
+# Travis-Check     : True
+# Script License   : Apache License, Version 2 or later
+# Maintainer       : Ankit Paraskar <Ankit.Paraskar@ibm.com>
+#
+# Disclaimer       : This script has been tested in root mode on given
+# ==========         platform using the mentioned version of the package.
+#                    It may not work as expected with newer versions of the
+#                    package and/or distribution. In such case, please
+#                    contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+# Variables
+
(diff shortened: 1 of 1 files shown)


m/mink-selenium2-driver/minkseleniumdriver_ubi8.5.sh
```diff
@@ -0,0 +1,60 @@
+#!/bin/bash -e
+
+# ----------------------------------------------------------------------------
+# Package          : minkseleniumdriver
+# Version          : latest
```
seth-priya, line 5: this should be the same as the version we are actually building in the script, v6.2.0 in this case?

m/mink-selenium2-driver/minkseleniumdriver_ubi8.5.sh
```diff
@@ -0,0 +1,60 @@
...
+#                    contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+# Variables
+
+PACKAGE_NAME=minkseleniumdriver
+PACKAGE_URL=https://github.com/minkphp/MinkSelenium2Driver.git
+#PACKAGE_VERSION is configurable can be passed as an argument.
+PACKAGE_VERSION=v6.2.0
```
seth-priya, line 26: are we actually using this?

m/mink-selenium2-driver/minkseleniumdriver_ubi8.5.sh
```diff
@@ -0,0 +1,60 @@
...
+    echo "$PACKAGE_URL $PACKAGE_NAME"
+    echo "$PACKAGE_NAME  |  $PACKAGE_URL |  $PACKAGE_VERSION | $OS_NAME | GitHub | Fail |  Clone_Fails"
+    exit 1
+fi
+
+
+
+cd "$HOME_DIR"/$PACKAGE_NAME || exit
+git fetch --tags
+latestTag=$(git describe --tags `git rev-list --tags --max-count=1`)
```
seth-priya, line 47: what is the version that was requested? looks like we are defining one but testing another one or am I missing something?

ankitpar: scripted to pick up the latest code from git repo but as i understand we have hardcoded dependency on specific version. Script is update accordingly also will modify other scripts

PR #2670, opened 2022-08-25
zap version v1.15.0 test and validate successfully

stutiibm:
## Checklist

### DIFF
--- a/z/zap/zap_v1.15.0.sh
+++ b/z/zap/zap_v1.15.0.sh
@@ -1,0 +1,80 @@
+#!/bin/bash -e
+# -----------------------------------------------------------------------------
+#
+# Package       : zap
+# Version       : v1.13.0, v1.19.1, v1.15.0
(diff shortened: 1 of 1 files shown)


seth-priya (commented):
@stutiibm are the build instructions same as that of the previously available script? if so we should update the existing script, if not, we should have only the current version specified here

stutiibm:
> @stutiibm are the build instructions same as that of the previously available script? if so we should update the existing script, if not, we should have only the current version specified here

Yes, I made the required changes in same script.

seth-priya:
But I see, two scripts zap_ubi_8.5.sh
 and zap_v1.15.0.sh, the latter being the one you are adding with this commit?

stutiibm:
> But I see, two scripts zap_ubi_8.5.sh and zap_v1.15.0.sh, the latter being the one you are adding with this commit?

Yes , Please merge zap_v1.15.0.sh

PR #2930, opened 2023-01-04
Currency: Added build-script and build_info for cxf

vinodk99:
## Checklist


kandarpamalipeddi:
@vinodk99 
Can you please explain why tests are skipped ?

CC: @nikhil-kalbande

kandarpamalipeddi:
> > @vinodk99 Can you please explain why tests were skipped ?
> > CC: @nikhil-kalbande
> 
> Tried to test in power and also in intel ,tests were parity so skipped the tests and also followed the community steps which mentioned to test ,attached below https://github.com/apache/cxf/blob/main/.github/workflows/codeql-analysis.yml

As far as I know, the given `mvn` command is to analyse the code.

@seth-priya , I feel it would be good if we provide build and test commands separately (probably comment the validation command). Also it is better if we attach Power and X86 logs to this PR for future reference. Any comments ?

CC: @nikhil-kalbande

seth-priya:
Yes, I suggested this for another one of these to Vinod - but same holds for this as well - we need to analyze the test failures even though they are in parity, we might be missing some dependency or config ot there might be a version mismatch or something else @kandarpamalipeddi . @vinodk99 please don't skip tests for any package unless otherwise agreed upon and make sure to segregate the build/test command as per @kandarpamalipeddi 's guidance

vinodk99:
The build and test logs are long, Travis check was failing due to  job exceeded the maximum log length, and has been terminated.Attached build and test logs for reference.
[cxf_logs.txt](https://github.com/ppc64le/build-scripts/files/10496630/cxf_logs.txt)

c/cxf/cxf_ubi_8.5.sh
```diff
@@ -1,4 +1,4 @@
-#!/bin/bash -e
+###!/bin/bash -e
```
asowani, line 1: Hi @vinodk99 - is this change intentional?

nikhil-kalbande: @kandarpamalipeddi - Can you please comments on it. Seems it is your commit.

asowani (approved):
Changes reviewed.

nikhil-kalbande (approved):

PR #2933, opened 2023-01-05
Jackson mod java8

stutiibm:
## Checklist


j/jackson-modules-java8/build-info.json
```diff
@@ -0,0 +1,14 @@
+{
+    "package_name" : "jackson-modules-java8",
+    "github_url":"https://github.com/FasterXML/jackson-modules-java8",
+    "version": "jackson-modules-java8-2.14.1",
+    "package_dir": "j/jackson-modules-java8",
+    "build_script": "jackson_module_java8_ubi_8_5.sh",
+    "validate_build_script": "true",
+    "use_non_root_user": "true",
```
kandarpamalipeddi, line 8: If `use_non_root_user` is set to`true`, build-script will be validated as non root used (sudo user).
But as per [this line](https://github.com/ppc64le/build-scripts/blob/89d614ba2c526b1ad51ad85a06e5facfd316bd94/j/jackson-modules-java8/jackson_module_java8_ubi_8_5.sh#L32), you are installing packages (using yum) without sudo.

I doubt, this would be gives troubles while validation. Please check and update accordingly.

If your intention to run it as non root user, please update the build-script (add sudo, where ever you need extended privileges) also update `Disclaimer` part of the script (Mention, script is tested as non-root user in Disclaimer).

kandarpamalipeddi (changes requested):

j/jackson-modules-java8/build-info.json
stutiibm: In that case it would be good if I make it "False(default-value)" as mine intension is not to run it as non root user.

kandarpamalipeddi (approved):

nikhil-kalbande (commented):
@stutiibm  In place of version pattern/regex you added script name only. Can you correct this in new PR.
cc : @kandarpamalipeddi 
image

PR #2934, opened 2023-01-06
Updated the directory related changes

Pratikt2312:
## Checklist


o/opentelemetry-go/build_info.json
```diff
@@ -2,14 +2,13 @@
...
     "docker_build": false,
     "build_script" : "opentelemetry-go_ubi_8.5.sh",
-    
+
     "v1.*.*" : {
- 
-	"args": {
-	    "GO_VERSION":"1.18.5"
+            "args": {
+            "GO_VERSION":"1.18.5"
```
nikhil-kalbande, line 11: @Pratikt2312 we will need to support older version also. If you keep version pattern `v1.*` only , it will not support older version `v0.13.0, v0.20.0` which are supported in current build scripts. Please handle both cases. 
You can keep two different entries as there is a change in GO version also.

o/opentelemetry-js/build_info.json
```diff
@@ -2,14 +2,14 @@
     "package_name" : "opentelemetry-js",
     "github_url":"https://github.com/open-telemetry/opentelemetry-js",
     "version": "v1.8.0",
-    "package_dir": "o/opentelemetry-js/latest",
+    "package_dir": "o/opentelemetry-js",
     "docker_build": false,
     "build_script" : "opentelemetry-js_ubi_8.5.sh",
-    
+
```
nikhil-kalbande, line 8: @Pratikt2312 Same comment as I mentioned for opentelemtry-go.
Here you can use different node version

o/opentelemetry-js/opentelemetry-js_ubi-8.5.sh
```diff
@@ -55,6 +55,15 @@ if ! npm run compile; then
 	exit 1
 fi
 
+# If api directory is available in the package then first need to run npm run compile in the api directory before running tests 
+# because it requires the ESM targets are generated with npm run compile in the API directory.
+  
+  if [ -d "api" ]; then
```
nikhil-kalbande, line 61: @Pratikt2312 Can you verify if update script works for earlier version `v1.4.0`
If not , please check new step is specific to v1.8.0 only?

o/opentelemetry-php/build_info.json
```diff
@@ -2,14 +2,13 @@
     "package_name" : "opentelemetry-php",
     "github_url":"https://github.com/open-telemetry/opentelemetry-php",
     "version": "0.0.17",
-    "package_dir": "o/opentelemetry-php/latest",
+    "package_dir": "o/opentelemetry-php",
     "docker_build": false,
     "build_script" : "opentelemetry-php_ubi_8.5.sh",
-    
+
```
nikhil-kalbande, line 8: @Pratikt2312 Same comments here. Please check

nikhil-kalbande (changes requested):

o/opentelemetry-js/opentelemetry-js_ubi-8.5.sh
Pratikt2312: Yes,I have verified the script again for version v1.4.0 and some other versions so its working fine.

nikhil-kalbande: Good. Then you just need to add support of older version in your build_info

o/opentelemetry-go/build_info.json
Pratikt2312: Yes I checked it for older versions like v0.13.0,v0.20.0,etc.and also made the required changes in json file.

o/opentelemetry-js/build_info.json
Pratikt2312: Here we use the same node version for latest version v1.8.0 and its working fine, so change is not required.

o/opentelemetry-js/opentelemetry-js_ubi-8.5.sh
Pratikt2312: Already have the support for older version in json file.

o/opentelemetry-php/build_info.json
Pratikt2312: Yes here also I checked the older build script version 0.0.12 on latest php version 8.0.25 its working well,so not need change.

nikhil-kalbande:
@Pratikt2312 is opening new PR's to separate out each package in different PR. So closing this

PR #2950, opened 2023-01-17
Antlr4

muskaan2199:
## Checklist

### DIFF
--- a/a/abyss/abyss_ubi_8.5.sh
+++ b/a/abyss/abyss_ubi_8.5.sh
@@ -1,0 +1,69 @@
+#!/bin/bash -e
+# -----------------------------------------------------------------------------
+#
+# Package	: abyss
+# Version	: v2.3.5
+# Source repo	: https://github.com/bcgsc/abyss.git
+# Tested on	: UBI: 8.5
+# Language      : Go 
+# Travis-Check  : True
+# Script License: Apache License, Version 2 or later
+# Maintainer	: Muskaan Sheik <Muskaan.Sheik@ibm.com>
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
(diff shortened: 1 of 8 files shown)


a/abyss/abyss_ubi_8.5.sh
```diff
@@ -0,0 +1,69 @@
...
+../configure --prefix=/usr/local/abyss
+
+if ! make; then
+        echo "Build fails"
+        exit 2
+fi
+
+if ! make install; then
+        echo "Install fails"
+        exit 2
```
nikhil-kalbande, line 60: @MuskaanSheik  we want keep return code "2" for test failure. So for build/install you can keep exit 1 only

a/abyss/abyss_ubi_8.5.sh
```diff
@@ -0,0 +1,69 @@
...
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+PACKAGE_NAME=abyss
+PACKAGE_VERSION=${1:-2.3.5}
+PACKAGE_URL=https://github.com/bcgsc/abyss.git
+
+yum install -y sudo
+sudo yum update -y
```
nikhil-kalbande, line 26: @MuskaanSheik  This script is supported for root or non-root user? why are we using sudo at some places only?

nikhil-kalbande (changes requested):

a/abyss/abyss_ubi_8.5.sh
muskaan2199: @nikhil-kalbande made the required change

a/abyss/abyss_ubi_8.5.sh
muskaan2199: @nikhil-kalbande It isn't needed. Removed that. I was following the steps in the abyss repository. But works the same without sudo so removed it

PR #2956, opened 2023-01-20
Updating build_info.json of currency packages, to include default branch info.

Ryan-Dmello:
## Checklist


nikhil-kalbande (approved):

nikhil-kalbande:
@Ryan-Dmello  We will need to add default_branch here https://github.com/ppc64le/build-scripts/blob/master/script/validate_builds.py#L104 as we expect it is mandatory field. This will help to avoid missing it in further PR's

Ryan-Dmello:
> @Ryan-Dmello We will need to add default_branch here https://github.com/ppc64le/build-scripts/blob/master/script/validate_builds.py#L104 as we expect it is mandatory field. This will help to avoid missing it in further PR's

Updated the build-info.json generation logic and added the requested params as mandatory

a/afero/build_info.json
```diff
@@ -2,6 +2,8 @@
     "package_name" : "afero",
     "github_url":"https://github.com/spf13/afero",
     "version": "v1.9.3",
+    "default_branch": "master",
```
seth-priya, line 5: Remind me - the reason why we need to set this in the build_info files rather than fetch it programmatically is because of the concern around the number of GitHub API calls, correct?

i/influxdb/build_info.json
```diff
@@ -2,6 +2,8 @@
     "package_name" : "influxdb",
     "github_url":"https://github.com/influxdata/influxdb.git",
     "version": "v2.5.1",
+    "default_branch": "master",
+    "build_script" : "influxdb_ubi_8.5.sh",
     "package_dir": "i/influxdb/Dockerfiles/",
```
seth-priya, line 7: I know this is not part of your changes, but just wondering, shouldn't the package_dir here be the top level directory "i/influxdb" as is the case for other packages?

a/afero/build_info.json
Ryan-Dmello: Correct.

i/influxdb/build_info.json
Ryan-Dmello: Yes, `package_dir` should be a top-level directory. But looks like in current case, it used for docker command.

a/apache-nifi/build_info.json
```diff
@@ -2,6 +2,8 @@
     "package_name" : "apache-nifi",
     "github_url":"https://github.com/apache/nifi",
     "version": "rel/nifi-1.17.0",
+    "default_branch": "main",
+    "build_script" : "apache-nifi_1.17.0_ubi_8.5.sh",
```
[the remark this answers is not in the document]
Ryan-Dmello: Line 17 is for any tags, related to generic release versions (branches matching `rel/nifi-`). In case we want to run for any branch line 6 script will be used.

seth-priya (approved):

PR #3133, opened 2023-04-19
fluent-bit v2.0.11 Script and Dockerfile

sumitd2:
## Checklist

- [X] Did you have **Legal approvals** for patch files ?


f/fluent-bit/fluent-bit-v2.0.11-ubi8.6.sh
```diff
@@ -0,0 +1,70 @@
...
+#
+# ----------------------------------------------------------------------------
+
+trap cleanup ERR
+function cleanup {
+  echo "ERROR: Script failed"
+}
+
+FLUENTBIT_VERSION=${1:-v2.0.11}
+USE_CENTOS_REPOS=${2:-1}
```
seth-priya, line 27: @sumitd2 why are we keeping it configurable? I understand that not everyone would want to enable centos repos, but then won't the build fail or are these packages available in RHEL/EPEL?

also looks like with these variable definitions, we are diverging a bit from our standard naming conventions of PACKAGE_NAME, PACKAGE_VERSION etc.

f/fluent-bit/fluent-bit-v2.0.11-ubi8.6.sh
```diff
@@ -0,0 +1,70 @@
...
+
+#Apply patch
+git apply $SCRIPT_PATH/fluent-bit-${FLUENTBIT_VERSION}.patch
+
+#Build
+cd $BUILD_HOME/fluent-bit/build/
+cmake -DFLB_TESTS_RUNTIME=On -DFLB_TESTS_INTERNAL=On -DFLB_RELEASE=On ..
+ret=0
+make -j $(nproc) || ret=$?
+if [ "$ret" -ne 0 ]
```
seth-priya, line 53: the "-e" on the first line should automatically take care of exiting on error or do we need this explicitly?

f/fluent-bit/fluent-bit-v2.0.11-ubi8.6.sh
```diff
@@ -0,0 +1,70 @@
...
+then
+	exit 1
+fi
+export FLUENTBIT_BIN=$BUILD_HOME/fluent-bit/build/bin/fluent-bit
+
+#Smoke test
+$FLUENTBIT_BIN --version
+
+#Test
+make test || true
```
seth-priya, line 63: this might not work well from a currency perspective as we would want to catch other unanalyzed failures on future versions, can we selectively disable the known failures?

seth-priya (commented):
@nikhil-kalbande please see if you have additional comments from a currency addition perspective

f/fluent-bit/fluent-bit-v2.0.11-ubi8.6.sh
nikhil-kalbande: yes, agree on this. 
Also if we capture build and test failure with different return code like exit 1 for build and exit 2 for test, will help.

f/fluent-bit/build_info.json
```diff
@@ -0,0 +1,16 @@
...
+    "package_dir": "f/fluent-bit",
+    "default_branch": "master",
+    "docker_cmd": "docker build -t ${package_name}:$PACKAGE_VERSION ${dir}",
+    "build_script": "fluent-bit-v2.0.11-ubi8.6.sh",
+    "docker_build": "true",
+    "validate_build_script": "true",
+
+    "2.0.11": {
+        "dir": "2.0.11_ubi_8",
+        "build_script": "fluent-bit-v2.0.11-ubi8.6.sh"
```
nikhil-kalbande, line 14: Can we make build script file name using separator `_` instead of  `-`. So it will be like fluent-bit_v2.0.11_ubi8.6.sh. Same for patch file. As I shared earlier `elasticsearch` is best example to refer for this - https://github.com/ppc64le/build-scripts/tree/master/e/elasticsearch

Also we have enabled ubi 8.7 container environment in travis check as well as currency jenkins, so if you can validate your script for 8.7 and include that in script file name.

nikhil-kalbande (changes requested):

f/fluent-bit/fluent-bit-v2.0.11-ubi8.6.sh
sumitd2: I am sure there are RHEL repos which have the required packages, so let the user enable them from outside the script. Variables definitions are fixed.

f/fluent-bit/fluent-bit-v2.0.11-ubi8.6.sh
sumitd2: We need to return the appropriate return codes for build and test failures, for currency. -e would just exit.

f/fluent-bit/fluent-bit-v2.0.11-ubi8.6.sh
sumitd2: Fixed.

f/fluent-bit/build_info.json
sumitd2: Done.

seth-priya:
Ok, I am fine with the changes/response, though will let @nikhil-kalbande confirm that the variable names etc. are all correct from currency perspective, I see use of "PKG_" instead of "PACKAGE_" as an example

nikhil-kalbande:
> Ok, I am fine with the changes/response, though will let @nikhil-kalbande confirm that the variable names etc. are all correct from currency perspective, I see use of "PKG_" instead of "PACKAGE_" as an example

Variables scope is just limited to shell script. But to keep it in parity with other scripts added to currency @sumitd2  please use PACKAGE instead of PKG, then I should be good on it.

sumitd2:
> > Ok, I am fine with the changes/response, though will let @nikhil-kalbande confirm that the variable names etc. are all correct from currency perspective, I see use of "PKG_" instead of "PACKAGE_" as an example
> 
> Variables scope is just limited to shell script. But to keep it in parity with other scripts added to currency @sumitd2 please use PACKAGE instead of PKG, then I should be good on it.

Done.

seth-priya (approved):

PR #3219, opened 2023-06-08
Currency:Added build script for reporting

Sunidhi-Gaonkar1:
## Checklist


seth-priya (approved):
Looks good to me. @nikhil-kalbande you might want to review this once from a currency perspective. @Sunidhi-Gaonkar1 confirmed that "gradlew build" runs both build and tests, do you still want her to try and split that into separate build/test stages (aka commands)?

seth-priya:
Ok, so one more thing as I look through the other scripts, it might make sense to put them in a dir/name them like o/opensearch-projects-reporting, to preserve the strong association with opensearch

nikhil-kalbande:
> Looks good to me. @nikhil-kalbande you might want to review this once from a currency perspective. @Sunidhi-Gaonkar1 confirmed that "gradlew build" runs both build and tests, do you still want her to try and split that into separate build/test stages (aka commands)?

yes @seth-priya . To make it follow standard template, we will like to have different stages unless there is any limitation.

r/reporting/reporting_ubi_8_7.sh
```diff
@@ -0,0 +1,42 @@
...
+# ---------------------------------------------------------------------------
+
+yum install -y  git gcc patch sudo make java-11-openjdk-devel python39 python39-devel bzip2-devel zlib-devel openssl-devel
+
+export PACKAGE_NAME=reporting
+export PACKAGE_URL=https://github.com/opensearch-project/reporting
+export PACKAGE_VERSION=${1:-2.8.0.0}
+
+useradd -G root -d /home/testuser testuser -p test123
+sudo -E -u testuser bash -c 'bash << EOF
```
nikhil-kalbande, line 28: @Sunidhi-Gaonkar1  I assume you need to be non root user to run this hence you are doing this way. 
But I will suggest to make "**use_non_root_user"= true** in build_info.json, and explicitly add sudo for yum command in build script and then you can make whole script run as normal user locally as well in currency setup.

nikhil-kalbande (changes requested):
Also as Priya suggested move you scripts to folder structure like **o/opensearch-reporting** and make related changes in build_info.json

r/reporting/reporting_ubi_8_7.sh
Sunidhi-Gaonkar1: Sure, I will make the required changes.

Sunidhi-Gaonkar1:
Attaching build and test logs: [Build_test_logs_reporting.txt](https://github.com/ppc64le/build-scripts/files/12547378/Build_test_logs_reporting.txt)

PR #3441, opened 2023-12-01
Currency:Updated build-script and build_info.json for Opentelemetry ruby 

vinodk99:
## Checklist

### DIFF
--- a/o/opentelemetry-ruby/opentelemetry-ruby_ubi_8.5.sh
+++ b/o/opentelemetry-ruby/opentelemetry-ruby_ubi_8.5.sh
@@ -2,7 +2,7 @@
 # -----------------------------------------------------------------------------
 #
 # Package	: opentelemetry-ruby
-# Version	: opentelemetry-sdk-experimental/v0.1.0
+# Version	: opentelemetry-sdk-experimental/v0.3.2
 # Source repo	: https://github.com/open-telemetry/opentelemetry-ruby.git
 # Tested on	: ubi 8.5
 # Language      : ruby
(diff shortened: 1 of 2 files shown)


o/opentelemetry-ruby/opentelemetry-ruby_ubi_8.5.sh
```diff
@@ -19,9 +19,9 @@
 # ----------------------------------------------------------------------------
 
 PACKAGE_NAME="opentelemetry-ruby"
-PACKAGE_VERSION=${1:-"opentelemetry-sdk-experimental/v0.1.0"}
```
nikhil-kalbande, line 22: @vinodk99 so this build script will not work for v0.1.0 now?. We need to add incremental support for new version, keeping support for old version also. So that through Currency we can run build for earlier supported version also.

o/opentelemetry-ruby/opentelemetry-ruby_ubi_8.5.sh
```diff
@@ -19,9 +19,9 @@
...
+PACKAGE_VERSION=${1:-"opentelemetry-sdk-experimental/v0.3.2"}
 PACKAGE_URL="https://github.com/open-telemetry/opentelemetry-ruby.git"
-export RUBY_VERSION=${RUBY_VERSION:-2.7.0}
```
nikhil-kalbande, line 24: @vinodk99 Check if we can make this available like GO version so we dont need to make hard coded value

nikhil-kalbande (changes requested):

o/opentelemetry-ruby/opentelemetry-ruby_ubi_8.5.sh
vinodk99: @nikhil-kalbande , it will work but it requires latest ruby version so i have updated the build-script with latest version

nikhil-kalbande (approved):

PR #3588, opened 2024-02-21
Adding scripts for multiple Katib/suggestion components 

pranavpandit1:
These changes are intended for multiple katib/suggestion components.

### DIFF
--- a/d/darts/LICENSE
+++ b/d/darts/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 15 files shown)


seth-priya:
So @pranavpandit1 couple of quick things
- we should have 1 script per GitHub URL, so if you have multiple components that map to the exact same GitHub tree, you should be creating only one script 
- we should have only 1 script in a single PR

pranavpandit1:
@seth-priya : I am okay with creating a single script, but from KF's perspective these are all different images.
kindly suggest.

seth-priya:
@pranavpandit1  lets stick to single script

pranavpandit1:
Closing this PR as all the components will be integrated in a single script. raising another PR.

PR #3589, opened 2024-02-21
Adding Script and Patched for multiple katib/suggestion components

pranavpandit1:
These changes targets to build multiple katib- suggestion components on PowerPC.


pranavpandit1:
@seth-priya  : as discussed, here is a single script for multiple components , kindly review.

seth-priya:
@pranavpandit1 just two things
- Travis check will have to be false as docker within docker is not supported
- Since the package name is now set to katib, the script and patches will need to go under k/katib and not s/suggestions
Request you to update this PR only and not raise a new one

pranavpandit1:
Done

Sapana-Khemkar:
@seth-priya is this still valid?

Sapana-Khemkar (approved):

PR #3752, opened 2024-04-25
Currency:Updated build-script,build_info.json and patch file for rules_rust

ramnathnayak:
## Checklist

- [x] Did you have **Legal approvals** for patch files ?


r/rules_rust/rules_rust_0.42.1.patch
```diff
@@ -0,0 +1,306 @@
+diff --git a/cargo/cargo_build_script_runner/bin.rs b/cargo/cargo_build_script_runner/bin.rs
+index 8730c809..54b8aa71 100644
+--- a/cargo/cargo_build_script_runner/bin.rs
++++ b/cargo/cargo_build_script_runner/bin.rs
```
nikhil-kalbande, line 4: so @vinodk99 will this patch changes work for older release version also or we should add different patch file per release version?

nikhil-kalbande (changes requested):

r/rules_rust/rules_rust_0.42.1.patch
vinodk99: @nikhil-kalbande, yes the patch file is working for older versions also

nikhil-kalbande (approved):

PR #4518, opened 2024-09-17
Milvus 2.4.11 Script, patch and Dockerfile

sumitd2:
## Checklist

- [X] Have you validated script on UBI 9 container
- [X] Did you have **Legal approvals** for patch files ?

### DIFF
--- a/m/milvus/milvus-v2.4.11-ubi9.3.sh
+++ b/m/milvus/milvus-v2.4.11-ubi9.3.sh
@@ -1,0 +1,203 @@
+#!/bin/bash -ex
+# ----------------------------------------------------------------------------
+#
+# Package       : milvus
+# Version       : 2.4.11
+# Source repo   : https://github.com/milvus-io/milvus
+# Tested on     : UBI 9.3 (docker)
+# Language      : C++, Go
+# Travis-Check  : True
+# Script License: Apache License, Version 2 or later
+# Maintainer    : Sumit Dubey <Sumit.Dubey2@ibm.com>
+#
+# Disclaimer: This script has been tested in root mode on given
+# ==========  platform using the mentioned version of the package.
+#             It may not work as expected with newer versions of the
+#             package and/or distribution. In such case, please
+#             contact "Maintainer" of this script.
+#
+# ----------------------------------------------------------------------------
+
+SCRIPT_PACKAGE_VERSION=v2.4.11
+PACKAGE_NAME=milvus
+PACKAGE_VERSION=${SCRIPT_PACKAGE_VERSION}
+PACKAGE_URL=https://github.com/milvus-io/${PACKAGE_NAME}
+CMAKE_VERSION=3.30.1
+CMAKE_REQUIRED_VERSION=3.30.1
+PYTHON_VERSION=3.10.2
+GO_VERSION=1.22.6
+ARROW_VERSION=16.0.0
+SCRIPT_PATH=$(dirname $(realpath $0))
+wdir=`pwd`
+
+for i in "$@"; do
+  case $i in
+    --power10)
+      APPLYMCPU=1
+      echo "Optimizing for Power 10"
+      shift
+      ;;
+    -*|--*)
+      echo "Unknown option $i"
+      exit 3
+      ;;
+    *)
+      PACKAGE_VERSION=$i
+      echo "Building ${PACKAGE_NAME} ${PACKAGE_VERSION}"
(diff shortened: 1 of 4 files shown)


m/milvus/milvus-v2.4.11-ubi9.3.sh
```diff
@@ -0,0 +1,203 @@
+#!/bin/bash -ex
+# ----------------------------------------------------------------------------
+#
+# Package       : milvus
+# Version       : 2.4.11
+# Source repo   : https://github.com/milvus-io/milvus
+# Tested on     : UBI 9.3 (docker)
```
seth-priya, line 7: Lets remove the (docker) string for consistency

m/milvus/milvus-v2.4.11-ubi9.3.sh
```diff
@@ -0,0 +1,203 @@
...
+  def package_info(self):
+    self.cpp_info.libs = tools.collect_libs(self)
+EOT
+}
+
+#Install centos and epel repos
+yum config-manager --add-repo https://mirror.stream.centos.org/9-stream/CRB/ppc64le/os
+yum config-manager --add-repo https://mirror.stream.centos.org/9-stream/AppStream//ppc64le/os
+yum config-manager --add-repo https://mirror.stream.centos.org/9-stream/BaseOS/ppc64le/os
+#rpm --import http://mirror.centos.org/centos/RPM-GPG-KEY-CentOS-Official
```
seth-priya, line 76: can we delete if not needed?

m/milvus/milvus-v2.4.11.patch
```diff
@@ -0,0 +1,435 @@
...
+-    image: minio/minio:RELEASE.2023-03-20T20-16-18Z
++    image: minio/minio:latest
+     ports:
+       - "9000:9000"
+       - "9001:9001"
+@@ -60,48 +56,6 @@ services:
+       timeout: 20s
+       retries: 3
+ 
+-  azurite:
```
seth-priya, line 51: what all is impacted by this?

seth-priya (commented):
in general, for any dependencies being install from source, we need to start putting some additional error handling to ensure that the version being built (for example for python , golang etc.) is not already installed, this is to support some upcoming changes in currency

m/milvus/milvus-v2.4.11-ubi9.3.sh
sumitd2: We have tested it using docker, so its a good idea to inform.

m/milvus/milvus-v2.4.11.patch
sumitd2: This is just for the golang tests.

m/milvus/milvus-v2.4.11-ubi9.3.sh
seth-priya: actually that's self implied when we say ubi 9.3

m/milvus/milvus-v2.4.11.patch
seth-priya: noted

m/milvus/milvus-v2.4.11-ubi9.3.sh
sumitd2: I meant it wasnt podman.

seth-priya: Okay, got it. Lets add  a comment in the script in that case, we don't want to complicate / make the text in these headers free form as currency / bulk search parses and uses the context from here.

sumitd2:
> in general, for any dependencies being install from source, we need to start putting some additional error handling to ensure that the version being built (for example for python , golang etc.) is not already installed, this is to support some upcoming changes in currency

Okay, but in that case script will have to be tested in that kind of an environment, which would open a can of worms. As you know we only test in a fresh docker container to keep things simple enough.

seth-priya:
> > in general, for any dependencies being install from source, we need to start putting some additional error handling to ensure that the version being built (for example for python , golang etc.) is not already installed, this is to support some upcoming changes in currency
> 
> Okay, but in that case script will have to be tested in that kind of an environment, which would open a can of worms. As you know we only test in a fresh docker container to keep things simple enough.

Yes, think of it as an extension built on top of the existing scripts and doing some pre/post processing.

seth-priya (approved):

PR #4609, opened 2024-10-11
Currency: Added build scripts and build_info.json files for pillow.

Salil4IBM:
pillow_ubi_8.10.sh - build pillow==10.3.0 .whl file(Maximo deliverable) with Python3.11, install .whl and test. Script to be run in UBI 8.10 container. Travis check set to False.
pillow_ubi_9.3.sh - build pillow==10.3.0 with Python3.11, install and test - part of Currency task. Script to be run in UBI 9.3 container. Travis check set to True.
build_info.json - build info for pillow_ubi_9.3.sh as per Currency process.
LICENCE - for the build scripts.

### DIFF
--- a/p/pillow/LICENSE
+++ b/p/pillow/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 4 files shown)


prachikurade1 (changes requested):
Support for latest version can be added - 10.4.0 for currency (ubi 9.3 script)

prachikurade1 (approved):
Approved

PR #4713, opened 2024-11-04
Currency: Added build scripts and build_info.json files for pyproj.

Salil4IBM:
pyproj_ubi_8.10.sh - build pyproj==3.6.1 .whl file with Python3.11, install .whl and test. Script to be run in UBI 8.10 container. Travis check set to False.
pyproj_ubi_9.3.sh - build pyproj==3.7.0 with Python3.11, install and test - part of Currency task. Script to be run in UBI 9.3 container. Travis check set to True.
build_info.json - build info for pyproj_ubi_9.3.sh as per Currency process.
LICENSE - for the build scripts.

### DIFF
--- a/p/pyproj/LICENSE
+++ b/p/pyproj/LICENSE
@@ -1,0 +1,201 @@
+                                 Apache License
+                           Version 2.0, January 2004
+                        http://www.apache.org/licenses/
+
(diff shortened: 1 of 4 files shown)


prachikurade1 (approved):
Approved

PR #5223, opened 2025-01-08
Currency: Added build_script and build_info.json for continuity

vinodk99:
Adding build_script and build_info.json

### DIFF
--- a/c/continuity/continuity_ubi_9.3.sh
+++ b/c/continuity/continuity_ubi_9.3.sh
@@ -1,0 +1,61 @@
+#!/bin/bash -e
+# -----------------------------------------------------------------------------
+#
+# Package          : continuity
+# Version          : v0.4.5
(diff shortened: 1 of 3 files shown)


c/continuity/continuity_ubi_9.3.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash -e
+# -----------------------------------------------------------------------------
+#
+# Package          : continuity
```
prachikurade1, line 4: Need to fix indentation for the header

prachikurade1 (changes requested):

c/continuity/continuity_ubi_9.3.sh
vinodk99: @prachikurade1 addressed review comments

prachikurade1 (approved):
Approved

PR #5450, opened 2025-02-19
Currency: Updated cassandra-driver_ubi_9.3.sh

ramnathnayak-ibm:
## Checklist

### DIFF
--- a/c/cassandra-driver/cassandra-driver_ubi_9.3.sh
+++ b/c/cassandra-driver/cassandra-driver_ubi_9.3.sh
@@ -65,7 +65,7 @@ if ! (python3 setup.py install) ; then
     exit 1
 fi
 
-if !(python3 -m tox -e gevent_loop); then
+if !(pytest tests/unit/ --disable-warnings --verbosity=2 --ignore=tests/unit/io/test_asyncioreactor.py --ignore=tests/unit/io/test_libevreactor.py --ignore=tests/unit/advanced/cloud/test_cloud.py --ignore=cassandra/datastax/cloud/__init__.py --ignore=tests/unit/test_host_connection_pool.py --ignore=tests/unit/io/test_twistedreactor.py); then
     echo "------------------$PACKAGE_NAME:install_success_but_test_fails---------------------"
     echo "$PACKAGE_URL $PACKAGE_NAME"
     echo "$PACKAGE_NAME  |  $PACKAGE_URL | $PACKAGE_VERSION | GitHub | Fail |  Install_success_but_test_Fails"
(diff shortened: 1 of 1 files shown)


c/cassandra-driver/cassandra-driver_ubi_9.3.sh
```diff
@@ -55,7 +55,7 @@ cd $PACKAGE_DIR
 git checkout $PACKAGE_VERSION
 
 #install necessary Python packages
-pip install wheel pytest tox nox
+pip install wheel pytest tox nox mock
 
 #Install
 if ! (python3 setup.py install) ; then
```
shubham-dayma-ibm, line 61: Try `pip install .` instead of `python3 setup.py install`, if it works keep that.

c/cassandra-driver/cassandra-driver_ubi_9.3.sh
```diff
@@ -65,7 +65,7 @@ if ! (python3 setup.py install) ; then
     exit 1
 fi
 
-if !(python3 -m tox -e gevent_loop); then
+if !(pytest tests/unit/ --disable-warnings --verbosity=2 --ignore=tests/unit/advanced/ --ignore=tests/unit/column_encryption/test_policies.py --ignore=tests/unit/io/test_eventletreactor.py --ignore=tests/unit/io/test_libevreactor.py --ignore=tests/unit/test_host_connection_pool.py); then
```
shubham-dayma-ibm, line 68: Make sure the skip tests are failing on amd64 as well. We cannot simply skip the tests if they are failing, we need to find resolution why they are failing and make necessary fixes. 
Take @vinodk99 help, if required. I suspect, apart from enabling port there shouldn't be any test case failure.

shubham-dayma-ibm (changes requested):

c/cassandra-driver/cassandra-driver_ubi_9.3.sh
vinodk99: > Make sure the skip tests are failing on amd64 as well. We cannot simply skip the tests if they are failing, we need to find resolution why they are failing and make necessary fixes. Take @vinodk99 help, if required. I suspect, apart from enabling port there shouldn't be any test case failure.

@shubham-dayma-ibm i have cross verified this script for the test part on x86 as well there are some tests which needs to be skipped due to port connection issues and also sent the updated test command to @ramnathnayak-ibm as well

c/cassandra-driver/cassandra-driver_ubi_9.3.sh
ramnathnayak-ibm: Included `pip install .` in the script.

c/cassandra-driver/cassandra-driver_ubi_9.3.sh
ramnathnayak-ibm: Thanks @vinodk99. @shubham-dayma-ibm I have changed the script accordingly and also have tested it locally.

shubham-dayma-ibm (approved):

PR #5588, opened 2025-03-17
Modified Go package version to fix CVE for Maximo.

Salil4IBM:
Fix CVE in stdlib package with higher Go version, as suggested by Grype tool.


prachikurade1 (approved):
Approved

PR #5725, opened 2025-03-27
P7zip 1705

Amitkumar282:
## Checklist


p/p7zip/p7zip_17.05_ubi9.3.sh
Amitkumar282, line 1: Approach behind this way of smoke test -
The reason --version or -version doesn't work with 7za is that p7zip does not implement a dedicated version flag like many other Linux utilities. Instead, running 7za without arguments prints its version and usage information by default.

./7za | head -n 2
It is reliable across versions of p7zip, since the version information is always in the first two lines as It directly extracts the version from the default output without unnecessary details.

sumitd2 (approved):

PR #6149, opened 2025-04-29
Fine-tune build script to make compatible with Currency build structure.

Salil4IBM:
Set paths relative to home directory in environment variables.


prachikurade1 (approved):
Approved

PR #6166, opened 2025-05-06
WA for Shapely wheel build issue due to pre-installed gcc 13 in wrapper script.

Salil4IBM:
Fix known issue with Shapely compilation using gcc-toolset-13(https://bugs.gentoo.org/915056 - found still occurs).
Note: Currently create_wheel_wrapper.sh pre-installs gcc-toolset-13 that fails Shapely wheel builds in python ecosystem.


prachikurade1 (approved):
Approved

PR #6306, opened 2025-06-01
Correcting dir structure in export commands

AaruniAggarwal:
Without these changes, torchaudio build is failing for both python 3.11 and 3.12 because of missing openblas and protobuf packages.
After making these changes, build is working fine both with and without wrapper script.


AaruniAggarwal:
[torchaudio-260-py311.pdf](https://github.com/user-attachments/files/20541966/torchaudio-260-py311.pdf)

shubham-dayma-ibm (approved):

PR #7173, opened 2025-07-13
Added build script for Pystan.

ramnathnayak-ibm:
## Checklist

### DIFF
--- a/p/pystan/pystan_ubi_9.3.sh
+++ b/p/pystan/pystan_ubi_9.3.sh
@@ -1,0 +1,99 @@
+#!/bin/bash -e
+# -----------------------------------------------------------------------------
+#
+# Package          : pystan
+# Version          : 3.10.0
+# Source repo      : https://github.com/stan-dev/pystan
+# Tested on        : UBI:9.3
+# Language         : Python
+# Travis-Check     : True
+# Script License   : Apache License, Version 2 or later
+# Maintainer       : Ramnath Nayak <Ramnath.Nayak@ibm.com>
+#
+# Disclaimer       : This script has been tested in root mode on given
+# ==========         platform using the mentioned version of the package.
+#                    It may not work as expected with newer versions of the
+#                    package and/or distribution. In such case, please
+#                    contact "Maintainer" of this script.
+#
+# ---------------------------------------------------------------------------
+
+# Variables
+PACKAGE_NAME=pystan
+PACKAGE_VERSION=${1:-3.10.0}
+PACKAGE_URL=https://github.com/stan-dev/pystan
+CURRENT_DIR=$(pwd)
+
+yum install -y git make wget gcc-toolset-13 openssl-devel python3.12 python3.12-pip python3.12-devel
+
+source /opt/rh/gcc-toolset-13/enable
+export PATH=/opt/rh/gcc-toolset-13/root/usr/bin:$PATH
+export LD_LIBRARY_PATH=/opt/rh/gcc-toolset-13/root/usr/lib64:$LD_LIBRARY_PATH
+
+# Clone and build CmdStan (for stanc)
+cd $CURRENT_DIR
+git clone https://github.com/stan-dev/cmdstan
+cd cmdstan
+git checkout v2.34.1
+git submodule update --init --recursive
+make build -j$(nproc)
+export PATH=$CURRENT_DIR/cmdstan/bin:$PATH
+which stanc
+stanc --version
+
+# Clone and build httpstan
+cd $CURRENT_DIR
+git clone https://github.com/stan-dev/httpstan
+cd httpstan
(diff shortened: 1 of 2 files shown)


p/pystan/pystan_ubi_9.3.sh
```diff
@@ -0,0 +1,99 @@
...
+# Clone and build httpstan
+cd $CURRENT_DIR
+git clone https://github.com/stan-dev/httpstan
+cd httpstan
+cp $CURRENT_DIR/cmdstan/bin/stanc $CURRENT_DIR/httpstan/httpstan/stanc
+chmod +x $CURRENT_DIR/httpstan/httpstan/stanc
+
+PYTHON_VER=$(python3.12 -c 'import sys; print(f"{sys.version_info.major}.{sys.version_info.minor}")')
+export PYTHON_PATH=$(which python${PYTHON_VER})
+ln -sf $PYTHON_PATH /usr/bin/python3
```
shubham-dayma-ibm, line 53: Remove symlink creation

p/pystan/pystan_ubi_9.3.sh
```diff
@@ -0,0 +1,99 @@
...
+cd $CURRENT_DIR
+
+# Clone the repository
+git clone $PACKAGE_URL
+cd $PACKAGE_NAME
+git checkout $PACKAGE_VERSION
+
+rm -f $CURRENT_DIR/httpstan/httpstan/stanc
+cp $CURRENT_DIR/cmdstan/bin/stanc $CURRENT_DIR/httpstan/httpstan/stanc
+chmod +x $CURRENT_DIR/httpstan/httpstan/stanc
```
shubham-dayma-ibm, line 77: Why are we removing and reperforming this? We already did same thing above

p/pystan/pystan_ubi_9.3.sh
```diff
@@ -0,0 +1,99 @@
...
+
+python3.12 -m pip install -r $CURRENT_DIR/httpstan/requirements.txt
+poetry build -v
+
+if ! python3.12 -m pip install -e . ; then
```
shubham-dayma-ibm, line 82: remove `-e`

p/pystan/pystan_ubi_9.3.sh
```diff
@@ -0,0 +1,99 @@
...
+export PATH=$CURRENT_DIR/cmdstan/bin:$PATH
+which stanc
+stanc --version
+
+# Clone and build httpstan
+cd $CURRENT_DIR
+git clone https://github.com/stan-dev/httpstan
+cd httpstan
+cp $CURRENT_DIR/cmdstan/bin/stanc $CURRENT_DIR/httpstan/httpstan/stanc
+chmod +x $CURRENT_DIR/httpstan/httpstan/stanc
```
shubham-dayma-ibm, line 49: What's the purpose copying and making it executable. We already installed it in my understanding `stanc --version`

shubham-dayma-ibm (changes requested):

p/pystan/pystan_ubi_9.3.sh
ramnathnayak-ibm: as they have set to python3.9 it will not work for other python version so i have added that step, so it will pick the installed python version and set that.
https://github.com/stan-dev/httpstan/blob/main/Makefile#L210

p/pystan/pystan_ubi_9.3.sh
ramnathnayak-ibm: We are not reperforming this, the above steps were performed in httpstan directory and at the bottom this is done for pystan directory. This is necessary because it expects a local, executable stanc within its repo—it doesn't use the global install, and will fail without it. This step ensures compatibility and avoids missing binary errors during build or runtime.

ramnathnayak-ibm: We're not re-performing the same step — the first stanc copy is for building httpstan, and the second is for pystan. Each expects a local, executable stanc ; they don’t use the global install, and since it may remove it during build, need to again copy that executable. This ensures it's present and usable where needed to avoid build/runtime failures. If this is not done we get a error stating same thing

p/pystan/pystan_ubi_9.3.sh
ramnathnayak-ibm: We're not re-performing the same step — the first stanc copy is for building httpstan, and the second is for pystan. Each expects a local, executable stanc ; they don’t use the global install, and since it may remove it during build, need to again copy that executable. This ensures it's present and usable where needed to avoid build/runtime failures. If this is not done we get a error stating same thing

p/pystan/pystan_ubi_9.3.sh
ramnathnayak-ibm: Removed.

shubham-dayma-ibm (approved):

PR #7650, opened 2025-12-28
Fix build script mapping for latest version i.e. 0.14.6

shivansh-ibm:
Updated build_info to map correct build script against latest version. Also pinned the numpy version to build wheel against the correct numpy version used in runtime


shivansh-ibm:
@shubham-dayma-ibm , please review this pr

shubham-dayma-ibm (approved):

PR #7910, opened 2026-02-13
Updated build-script for vllm v0.11.1

pooja0805:
## Checklist


v/vllm/vllm_v0.11.1_ubi_9.5.sh
```diff
@@ -77,9 +73,7 @@ yum install -y \
   libtool \
   pkgconfig \
   make \
-  gcc
-
-$PYTHON --version
+  gcc openblas-devel
```
cdeepali: @pooja0805 Do we use openblas-devel? or build openblas from source in our scripts?

pooja0805: We require the openblas-devel dependency as well, or else the torch build fails. We also download the OpenBLAS wheel.

cdeepali (approved):

PR #7935, opened 2026-02-19
feat(CI): add large runner support for currency workflows

anup-kodlekere:
This PR builds on #7926 by allowing `currency-build.yml` to also use larger runners. 

By default, if the `large-runner-label` field is empty, the build will default to the `ubuntu-24.04-ppc64le-p10` runner. 
Incase the build fails with the default runner, maintainers now will be able to trigger another build, this time with a larger runner. 

For builds which are rerun on a larger runner, the Workflow Run will contain the appropriate title.

When triggered normally:
image

When re-triggered on large runner
image

### DIFF
--- a/.github/workflows/currency-build.yaml
+++ b/.github/workflows/currency-build.yaml
@@ -41,7 +41,22 @@ on:
         required: false
         default: 'None'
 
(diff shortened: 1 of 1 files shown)


ghatwala:
@Vipul-Ajmera  @nikhil-kalbande  @Alston-Dias - please check and review above feature code

PR #8350, opened 2026-07-06
Pr build yaml enhancement

stutiibm:
## Checklist

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -28,32 +28,78 @@ run-name: >
 
 jobs:
 
-  build_info:
+  check_changes:
     runs-on: ${{ github.event_name == 'pull_request' && 'ubuntu-24.04-ppc64le-p10' || inputs.large-runner }}
+    outputs:
+      should_build: ${{ steps.filter.outputs.should_build }}
+      changed_files: ${{ steps.filter.outputs.changed_files }}
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
+
+      - name: Fix workspace permissions
+        run: sudo chown -R $USER:$USER .
+
+      - name: Check for relevant file changes
+        id: filter
+        run: |
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
+            echo "should_build=true" >> $GITHUB_OUTPUT
+            echo "Workflow dispatch - proceeding with build"
+            exit 0
+          fi
+
+          # Fetch base branch with full history
+          git fetch origin ${{ github.base_ref }}
+          CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD)
+
+          echo "Changed files:"
+          echo "$CHANGED_FILES"
+
+          # Store changed files for reuse in later jobs
+          echo "changed_files<<EOF" >> $GITHUB_OUTPUT
+          echo "$CHANGED_FILES" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+          # Check for build_info.json, .sh scripts (excluding specific directories), or Dockerfile
+          RELEVANT_CHANGES=$(echo "$CHANGED_FILES" | grep -E '(build_info\.json|\.sh$|Dockerfile)' | grep -vE '^(gha-script|process_bom|script|templates|travis-currency-ymls|travis-ymls)/' || true)
+
+          if [ -n "$RELEVANT_CHANGES" ]; then
+            echo "should_build=true" >> $GITHUB_OUTPUT
+            echo "✅ Found relevant changes:"
+            echo "$RELEVANT_CHANGES"
+          else
+            echo "should_build=false" >> $GITHUB_OUTPUT
+            echo "⏭️ Skipping PR build CI check - no changes related to build_info.json, build scripts (.sh), or Dockerfile"
+          fi
+
+  build_info:
+    needs: check_changes
+    if: needs.check_changes.outputs.should_build == 'true'
+    runs-on: ${{ github.event_name == 'pull_request' && 'ubuntu-24.04-ppc64le-p10' || inputs.large-runner }}
+    outputs:
+      wheel_build_enabled: ${{ steps.set_flags.outputs.wheel_build_enabled }}
+      has_sh_changes: ${{ steps.set_flags.outputs.has_sh_changes }}
+      has_dockerfile_changes: ${{ steps.set_flags.outputs.has_dockerfile_changes }}
+      docker_build_enabled: ${{ steps.set_flags.outputs.docker_build_enabled }}
+      build_package_enabled: ${{ steps.set_flags.outputs.build_package_enabled }}
 
     steps:
       - name: Checkout code (Pull Request)
         if: github.event_name == 'pull_request'
         uses: actions/checkout@v6
         with:
-             fetch-depth: 0
-             ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: 0
+          ref: ${{ github.event.pull_request.head.sha }}
 
       - name: Checkout code (Workflow Dispatch)
         if: github.event_name == 'workflow_dispatch'
         uses: actions/checkout@v6
         with:
           ref: refs/pull/${{ inputs.pr_number }}/head
-
-      - name: Install required packages
-        run: |
-          sudo apt update -y
-          sudo apt-get install -y file jq
-
-      - name: Install Python dependencies
-        run: |
-          pip3 install --force-reinstall -v "requests==2.31.0"
-          pip3 install --upgrade docker
 
       - name: Debug - Check user and workspace ownership
         run: |
@@ -69,6 +115,15 @@ jobs:
       - name: Fix workspace permissions
         run: sudo chown -R $USER:$USER .
 
+      - name: Install required packages
+        run: |
+          sudo apt update -y
+          sudo apt-get install -y file jq
+
+      - name: Install Python dependencies
+        run: |
+          pip3 install --force-reinstall -v "requests==2.31.0"
+          pip3 install --upgrade docker
 
       - name: Set PR number
         run: echo "PR_NUMBER=${{ github.event.pull_request.number || inputs.pr_number }}" >> $GITHUB_ENV
@@ -92,8 +147,14 @@ jobs:
 
       - name: Locate and parse build_info.json
         run: |
+          # Reuse changed files from check_changes job
+          CHANGED_FILES="${{ needs.check_changes.outputs.changed_files }}"
 
-          CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD)
+          # If workflow_dispatch, check_changes did not compute changed_files — fetch them now
+          if [ "${{ github.event_name }}" == "workflow_dispatch" ]; then
+            git fetch origin ${{ github.base_ref }} --depth=1
+            CHANGED_FILES=$(git diff --name-only origin/${{ github.base_ref }}...HEAD)
+          fi
 
           BUILD_INFO_FILE=$(echo "$CHANGED_FILES" | grep 'build_info.json' | head -n 1)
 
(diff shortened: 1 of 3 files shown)


stutiibm:
## Add pre-flight job-control enhancements to `pr-build.yaml`

### Background

The current `pr-build.yaml` has no gating logic — every PR unconditionally allocates a `ppc64le` runner and runs all jobs regardless of what files were actually changed. This wastes self-hosted runner time on PRs that only touch documentation, templates, or unrelated config files. Additionally, the wheel and docker jobs use `success()` as their `if` condition, which is semantically incorrect — it runs those jobs even when the package does not support wheel builds or when no Dockerfile was changed.

---

### Changes Proposed

#### 1. Add `check_changes` pre-flight job

A new first job `check_changes` runs before anything else. It examines which files changed in the PR and sets two outputs:

- `should_build` — `true` if any relevant file changed (`build_info.json`, `*.sh` scripts, `Dockerfile`), `false` otherwise
- `changed_files` — the full list of changed files, stored as a job output for reuse by downstream jobs

Files inside the following directories are explicitly excluded from the relevance check even if they are `.sh` files, since they are tooling/infra scripts and not package build scripts:
`gha-script/`, `process_bom/`, `script/`, `templates/`, `travis-currency-ymls/`, `travis-ymls/`

On `workflow_dispatch`, the job fast-paths to `should_build=true` without running `git diff`.

**Why:** Avoids allocating any `ppc64le` runner at all for doc-only or config-only PRs.

---

#### 2. Gate `build_info` on `check_changes` output

`build_info` now declares `needs: check_changes` and adds:

```
if: needs.check_changes.outputs.should_build == 'true'
```

**Why:** The `ppc64le` self-hosted runner is only allocated when `check_changes` confirms something relevant changed.

---

#### 3. Add job-control flag outputs to `build_info`

A new `Set job control flags` step (id: `set_flags`) is added to `build_info`. It reads `build_info.json` and cross-references it with the list of changed files to compute five boolean outputs:

| Output | Condition |
|---|---|
| `build_package_enabled` | `build_info.json` changed OR a `.sh` script inside `PACKAGE_DIR` changed |
| `wheel_build_enabled` | `wheel_build: true` in `build_info.json` AND a `.sh` script changed |
| `docker_build_enabled` | `docker_build: true` in `build_info.json` AND a `Dockerfile` changed |
| `has_sh_changes` | Any `.sh` script changed inside `PACKAGE_DIR` |
| `has_dockerfile_changes` | Any `Dockerfile` changed |

Each flag is computed with explicit log output (✅ / ⏭️) so the intent is visible in the job logs.

**Why:** Centralises all build-gating decisions in one place. Downstream jobs simply check the flag — no inline shell logic needed to decide whether to run.

---

#### 4. Reuse `changed_files` from `check_changes` in `build_info`

The `Locate and parse build_info.json` step previously re-ran `git diff` from scratch. It now reads the output directly:

```
CHANGED_FILES="${{ needs.check_changes.outputs.changed_files }}"
```

A fallback `git diff` is only run when the event is `workflow_dispatch`, because in that case `check_changes` fast-pathed without storing `changed_files`.

**Why:** Eliminates a redundant `git diff` network call. The data was already computed in `check_changes`.

---

#### 5. Gate `build` job on `build_package_enabled` flag

The `build` job adds:

```
if: needs.build_info.outputs.build_package_enabled == 'true'
```

Previously the `build` job ran unconditionally after `build_info` succeeded — meaning it allocated a runner even when only a `Dockerfile` changed (which should only trigger `build_docker`, not `build`).

**Why:** Avoids wasting a runner on a build job that has nothing to build.

---

#### 6. Replace `build_package.sh` with `execute_changed_scripts.py` in the `build` job

The `Build Package` step previously called `bash ./gha-script/build_package.sh` which ran a single hardcoded script. It is now replaced with:

```
python3 ./gha-script/execute_changed_scripts.py
```

`execute_changed_scripts.py` is a new script that:
1. Reads the `CHANGED_FILES` environment variable
2. Filters for `.sh` scripts only
3. For each changed script, parses its header metadata (`# Version`, `# Ci-Check`, `# Tested on`, `# Use Non-Root User`)
4. Skips scripts where `Ci-Check` is not `true` or where no version is found in the header
5. Sets per-script environment variables (`PKG_DIR_PATH`, `BUILD_SCRIPT`, `VERSION`, `TESTED_ON`, `NON_ROOT_BUILD`)
6. Calls `bash gha-script/build_package.sh` for each qualifying script individually
7. Fails fast with the script's exit code if any execution fails
8. Prints an execution summary at the end (Executed / Skipped / Total)

**Why:** The old approach ran a single blanket build. This approach runs only the scripts that were actually touched in the PR, respects per-script CI flags, and supports multiple changed scripts in a single PR.

As a paired change, `validate_builds.py` in `build_info` no longer calls `trigger_script_validation_checks()` (which spawned a Docker container to run the script). It now only performs static validation — header checks, license file presence, CRLF line ending detection. Actual execution is fully delegated to the `build` job. This eliminates duplicate execution.

---

#### 7. Fix wheel job `if` conditions — replace `success()` with flag check

All wheel jobs (`wheel_build_py310` through `wheel_build_py314`) previously used:

```
if: ${{ success() }}
```

This is semantically wrong — `success()` only checks that the parent job didn't fail. It does not check whether a wheel build is warranted. All wheel jobs now use:

```
if: needs.build_info.outputs.wheel_build_enabled == 'true'
```

**Why:** Previously, any PR that passed `build_info` would trigger all 4 wheel jobs regardless of whether `wheel_build: true` was set in `build_info.json` or whether any `.sh` script was even changed. This wasted 4 ppc64le runner allocations per PR.

---

#### 8. Fix `build_docker` `if` condition — same `success()` problem

`build_docker` previously used `if: ${{ success() }}` and relied on internal shell `if` checks inside the step script to bail out. It now uses:

```
if: needs.build_info.outputs.docker_build_enabled == 'true'
```

**Why:** Same as above — the job no longer allocates a runner at all if a Docker build is not warranted.

---


### Job Flow After This Change

```
PR opened
    │
    ▼
check_changes  (ppc64le runner)
    │  should_build=false → all jobs skipped, no more runners allocated
    │  should_build=true  ↓
    ▼
build_info  (ppc64le runner)
    │  computes: build_package_enabled, wheel_build_enabled, docker_build_enabled
    │
    ├─► build               (if build_package_enabled=true)
    ├─► wheel_build_py310   (if wheel_build_enabled=true)
    ├─► wheel_build_py311   (if wheel_build_enabled=true)
    ├─► wheel_build_py312   (if wheel_build_enabled=true)
    ├─► wheel_build_py313   (if wheel_build_enabled=true)
    ├─► wheel_build_py314   (if wheel_build_enabled=true)
    └─► build_docker        (if docker_build_enabled=true)
```

gha-script/execute_changed_scripts.py
```diff
@@ -0,0 +1,158 @@
...
+            continue
+
+        # Set environment variables for build_package.sh
+        os.environ['PKG_DIR_PATH'] = os.path.dirname(script_path) + '/'
+        os.environ['BUILD_SCRIPT'] = os.path.basename(script_path)
+        os.environ['VERSION'] = version
+        os.environ['TESTED_ON'] = tested_on
+        os.environ['NON_ROOT_BUILD'] = use_non_root
+
+        print(f"\n🚀 Executing: {script_path} with version {version}")
```
cdeepali, line 130: All these logging statements are not being printed until the actual build has completed. 

Invoke script in unbuffered mode using:

 `python3  -u ./gha-script/execute_changed_scripts.py`

gha-script/execute_changed_scripts.py
```diff
@@ -0,0 +1,158 @@
...
+        print(f"   Command: bash gha-script/build_package.sh")
+
+        # Execute build_package.sh which will run the script
+        raw_exit = os.system('bash gha-script/build_package.sh')
+        # os.system() returns waitpid status — shift right 8 bits to get the real exit code
+        # Without this, exit code 256 wraps to 0 via sys.exit() and the job shows green on failure
+        exit_code = raw_exit >> 8
```
cdeepali: Change to os.waitstatus_to_exitcode instead of shifting the bits of the code.

.github/workflows/pr-build.yaml
```diff
@@ -28,33 +28,79 @@ run-name: >
...
+          echo "Changed files:"
+          echo "$CHANGED_FILES"
+
+          # Store changed files for reuse in later jobs
+          echo "changed_files<<EOF" >> $GITHUB_OUTPUT
+          echo "$CHANGED_FILES" >> $GITHUB_OUTPUT
+          echo "EOF" >> $GITHUB_OUTPUT
+
+          # Check for build_info.json, .sh scripts (excluding specific directories), or Dockerfile
+          RELEVANT_CHANGES=$(echo "$CHANGED_FILES" | grep -E '(build_info\.json|\.sh$|Dockerfile)' | grep -vE '^(gha-script|process_bom|script|templates|travis-currency-ymls|travis-ymls)/' || true)
```
cdeepali, line 65: Should RELEVANT_CHANGES be printed to GITHUB_OUTPUT as CHANGED_FILES, because this is then later used in execute_changed_scripts.py for actual build?

gha-script/validate_builds.py
```diff
@@ -315,14 +315,20 @@ def trigger_build_validation_ci(pr_number):
             # perform basic validation check
             trigger_basic_validation_checks(file_name)
 
-            #check ci-check from package header  
+            #check ci-check from package header
             ci_check=package_data['ci_check'].lower()
```
cdeepali: Is Ln 319 required?

.github/workflows/pr-build.yaml
```diff
@@ -28,33 +28,79 @@ run-name: >
 
 jobs:
 
+  check_changes:
+    runs-on: ${{ github.event_name == 'pull_request' && 'ubuntu-24.04-ppc64le-p10' || inputs.large-runner }}
```
cdeepali: Can this run on ubuntu-latest instead of ppc runner.

.github/workflows/pr-build.yaml
```diff
@@ -122,13 +183,74 @@ jobs:
...
       - name: Read build_info.json
         run: |
           sudo chown -R $USER:$USER .
           chmod +x ./gha-script/read_buildinfo.sh
           bash ./gha-script/read_buildinfo.sh
 
+      - name: Set job control flags
+        id: set_flags
+        run: |
+          PACKAGE_DIR=$(jq -r '.package_dir // ""' $BUILD_INFO_FILE)
```
cdeepali, line 204: Add validation for BUILD_INFO_FILE.

gha-script/validate_builds.py
```diff
@@ -315,14 +315,20 @@ def trigger_build_validation_ci(pr_number):
...
-            if ci_check=="true":
 
-                # Build/test script files
-                trigger_script_validation_checks(file_name)
-            else:
-                print("Skipping Build script validation for {} as CI-Check flag is set to False".format(file_name))
+            # COMMENTED OUT: Script execution moved to build job via execute_changed_scripts.py
+            # This avoids duplicate execution - validation happens here, actual build happens in build job
+            # if ci_check=="true":
```
cdeepali: Can remove this commented code.

gha-script/execute_changed_scripts.py
stutiibm: Done. Added -u flag to the python3 invocation in the build job (python3 -u ./gha-script/execute_changed_scripts.py). This forces unbuffered stdout/stderr so build logs stream in real time to the GitHub Actions UI rather than being held in the buffer and dumped all at once at the end.

gha-script/execute_changed_scripts.py
stutiibm: Done. Replaced the raw os.system() return value handling with os.waitstatus_to_exitcode() in [execute_changed_scripts.py](https://github.com/mcasapi/McasPoliciesEndPoint/v2?policy=NativeApps&redir=0&McasSrc=vscode-webview%3A%2F%2F0ae4q6dueaaj172qah1vf3shu3kdgehu7q0379h3hnvnopv7bnat%2Fgha-script%2Fexecute_changed_scripts.py&McasId=9137feaf-8257-4ffa-b2a9-df18acfd6bc2&McasHash=33ca12a68c6cfa50ea23c69a34a6c2cd98e2a7b711ec9264ffcd86d7e4494e42).

The problem with using >> 8 (or ignoring the raw value entirely) is that os.system() returns a waitpid-style status, not a plain exit code — on some platforms a non-zero exit like 1 gets encoded as 256, and a naive check or bit-shift can mask real failures. os.waitstatus_to_exitcode() is the correct POSIX-aware way to decode it and handles all cases (normal exit, signal termination, etc.).

.github/workflows/pr-build.yaml
stutiibm: No change needed here — this is intentional. changed_files output carries the full list of changed files (all git diff output) so that later jobs (build_info, build) can reuse it without re-running git diff. RELEVANT_CHANGES is only a filtered subset used internally within check_changes to decide whether to set should_build=true/false. Passing the filtered list downstream would break the build_info job's ability to locate build_info.json and detect .sh/Dockerfile changes across the full changeset.

gha-script/validate_builds.py
stutiibm: Done. Removed the unused ci_check = package_data['ci_check'].lower() line and the large commented-out trigger_script_validation_checks block from trigger_build_validation_ci() in [validate_builds.py](https://github.com/mcasapi/McasPoliciesEndPoint/v2?policy=NativeApps&redir=0&McasSrc=vscode-webview%3A%2F%2F0ae4q6dueaaj172qah1vf3shu3kdgehu7q0379h3hnvnopv7bnat%2Fgha-script%2Fvalidate_builds.py&McasId=e7667577-1987-44b2-8133-a10b914070ee&McasHash=70331fcd66da56f583e85fb26750241e1de66483cefba6b20fa0e9151626fa06). The comment block was left over from when script execution lived inside validate_builds.py — that work has now moved to execute_changed_scripts.py in the build job, so the dead code served no purpose and was just noise.

.github/workflows/pr-build.yaml
stutiibm: Done. Changed check_changes runner from ubuntu-24.04-ppc64le-p10 to ubuntu-latest. This job only runs git diff and bash string matching — it has no ppc64le-specific work, so there's no reason to consume a self-hosted runner slot for it. Also removed the sudo chown step that was in that job, since GitHub-hosted runners always start with a fresh workspace with correct ownership. The ppc64le runner is now only allocated when actual build/validation work starts in build_info.

.github/workflows/pr-build.yaml
stutiibm: Done. Added an explicit guard at the top of the set_flags step that checks both:

BUILD_INFO_FILE env var is non-empty (set by the preceding "Locate and parse build_info.json" step)
The file actually exists on disk
Without this guard, a misconfigured or missing build_info.json would cause all the jq calls to fail with a generic parse error, making the root cause unclear. With the guard, the step now exits immediately with a descriptive message pointing exactly at the problem.

gha-script/validate_builds.py
stutiibm: Yes, removed commented part

stutiibm:
Addressed review-comments on PR and tested changes too. 
Information regarding tests is mention in issue here : https://github.ibm.com/ppc64le-automation/CurrencyAutomation/issues/3961#issuecomment-222078530

.github/workflows/pr-build.yaml
cdeepali: Okay, one question, at line 214 in pr-build.yml 

SH_SCRIPT_CHANGED=$(echo "$CHANGED_FILES" | grep -E "^$PACKAGE_DIR/.*\.sh$" || true)

will this also pick up script file changes from the folder that are skipped while detectign relevant changes - example will it also try to build if script under gha-script directory is updated?

stutiibm: Line 214 is safe — it uses $PACKAGE_DIR as a prefix (e.g. p/pandas/), so gha-script/ files can never match it.

But there was a real bug nearby at line 161 — when build_info.json wasn't directly changed, the code picked the first file from $CHANGED_FILES to figure out which package directory to look in. If a gha-script/ file happened to appear first in the list, it would try to find gha-script/.../build_info.json and fail.

Fix: filter out the excluded directories first, then take the first file — same exclusion list already used in check_changes.

cdeepali: Thanks Stuti.

cdeepali (approved):

PR #8406, opened 2026-07-14
Currency build yaml changes licenses extract script

stutiibm:
## Checklist

### DIFF
--- a/.github/workflows/currency-build.yaml
+++ b/.github/workflows/currency-build.yaml
@@ -3,7 +3,6 @@ on:
   workflow_dispatch:
     branches:
       - master
-      - replica-master
 
     inputs:
       package_name:
@@ -117,7 +116,45 @@ jobs:
         with:
           name: package-cache
           path: package-cache.tar.gz
-                
+
+  # ---------------------------------------------------------------------------
+  # install_scan_tools: Installs grype ONCE (static Go binary — fully portable)
+  # and caches it as an artifact for all wheel_build_pyXXX jobs.
+  # Scancode-toolkit is installed fresh per wheel_build job using pip cache
+  # (Option 1) — pip downloads are cached so reinstall is fast.
+  # ---------------------------------------------------------------------------
+  install_scan_tools:
+    needs: build_info
+    if: ${{ inputs.wheel_build == 'true' }}
+    runs-on: ubuntu-24.04-ppc64le-p10
+    steps:
+      - name: Install system dependencies
+        run: |
+          sudo apt update -y
+          sudo apt install -y wget curl tar jq
+
+      - name: Install grype
+        run: |
+          GRYPE_VERSION=$(curl -s --retry 3 https://api.github.com/repos/anchore/grype/releases/latest | jq -r '.tag_name')
+          echo "Installing grype $GRYPE_VERSION for ppc64le..."
+          wget --tries=3 https://github.com/anchore/grype/releases/download/$GRYPE_VERSION/grype_${GRYPE_VERSION#v}_linux_ppc64le.tar.gz
+          tar -xzf grype_${GRYPE_VERSION#v}_linux_ppc64le.tar.gz
+          mkdir -p scan-tools-bin
+          mv grype scan-tools-bin/grype
+          chmod +x scan-tools-bin/grype
+          echo "Grype version:"
+          scan-tools-bin/grype version
+
+      - name: Archive grype binary
+        run: tar -czf scan-tools-cache.tar.gz scan-tools-bin/
+
+      - name: Upload scan-tools-cache artifact
+        uses: actions/upload-artifact@v6
+        with:
+          name: scan-tools-cache
+          path: scan-tools-cache.tar.gz
+
+        
   build:
     needs: build_info
     if: ${{ inputs.validate_build_script == 'true' }}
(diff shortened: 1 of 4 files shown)


stutiibm:
## What this PR does

1. **Grype installed once, not 5 times**
   Added a new `install_scan_tools` job that installs the grype binary once
   and caches it as a GHA artifact. All 5 wheel build jobs reuse it.

2. **Scancode install made faster with pip cache**
   Scancode is still installed per wheel build job (it's a Python script, not
   portable across runners), but a pip download cache is added so after the
   first run, reinstall takes ~1-2 min instead of ~10 min.

3. **Wheel only uploads to COS after scans pass**
   Each wheel build job now runs scancode and grype scans before uploading
   the wheel to COS. If either scan fails, the wheel upload is skipped.
   This replaces the old standalone wheel_licenses job.

4. **New wheel_license_extract job**
   After all wheel builds complete, this job downloads the scancode JSON
   outputs from COS and runs licenses_extract_script.py on each one to
   extract SPDX license identifiers.

5. **Updated scan scripts**
   - scancode_wheel_scan.sh — scans wheel from workspace root instead of
     package-cache/wheels/
   - grype_wheel_scan.sh — uses cached grype binary ($GRYPE_BIN), scans
     from workspace root

Note: This PR depends on the licenses_extract_script.py file being merged : [8402](https://github.com.mcas.ms/ppc64le/build-scripts/pull/8402)

.github/workflows/currency-build.yaml
```diff
@@ -117,7 +116,45 @@ jobs:
...
+        run: |
+          sudo apt update -y
+          sudo apt install -y wget curl tar jq
+
+      - name: Install grype
+        run: |
+          GRYPE_VERSION=$(curl -s --retry 3 https://api.github.com/repos/anchore/grype/releases/latest | jq -r '.tag_name')
+          echo "Installing grype $GRYPE_VERSION for ppc64le..."
+          wget --tries=3 https://github.com/anchore/grype/releases/download/$GRYPE_VERSION/grype_${GRYPE_VERSION#v}_linux_ppc64le.tar.gz
+          tar -xzf grype_${GRYPE_VERSION#v}_linux_ppc64le.tar.gz
```
cdeepali: Verify checksum for the archive before extracting it.

gha-script/scanner-scripts/scancode_wheel_scan.sh
```diff
@@ -7,15 +7,15 @@ echo "----------Installing dependencies -----------------"
 sudo apt update -y && sudo apt install -y file git python3.12 python3.12-venv python3-pip python3.12-dev build-essential unzip patch wget tar libffi-dev zlib1g-dev libssl-dev libxml2-dev libxslt1-dev libicu-dev pkg-config
 
 echo "----------Installed dependencies -----------------"
+# pip download cache (~/.cache/pip) is restored by actions/cache in the workflow
+# so this git clone + pip install is fast after the first run
 git clone https://github.com/nexB/scancode-toolkit.git
 cd scancode-toolkit
 git checkout v32.4.0
```
cdeepali: Use a variable for scancode version, so that it can be easily updated later.

gha-script/scanner-scripts/scancode_wheel_scan.sh
```diff
@@ -7,15 +7,15 @@ echo "----------Installing dependencies -----------------"
 sudo apt update -y && sudo apt install -y file git python3.12 python3.12-venv python3-pip python3.12-dev build-essential unzip patch wget tar libffi-dev zlib1g-dev libssl-dev libxml2-dev libxslt1-dev libicu-dev pkg-config
 
 echo "----------Installed dependencies -----------------"
+# pip download cache (~/.cache/pip) is restored by actions/cache in the workflow
+# so this git clone + pip install is fast after the first run
 git clone https://github.com/nexB/scancode-toolkit.git
```
cdeepali: Can we save cache this? currently it is being cloned for each job.

.github/workflows/currency-build.yaml
stutiibm: Done. In the install_scan_tools job, after downloading the grype archive, we now also download the _checksums.txt file from the same GitHub release and verify using grep ... | sha256sum --check --status before running tar. The job exits with [ERROR] if verification fails. This follows the same pattern used for Trivy in the existing scan scripts.

gha-script/scanner-scripts/scancode_wheel_scan.sh
stutiibm: Done. Added SCANCODE_VERSION="v32.4.0" as a variable at the top of scancode_wheel_scan.sh. The hardcoded git checkout v32.4.0 is now replaced with git checkout $SCANCODE_VERSION. The same variable is also used in the new fetch_scancode_toolkit.sh script, so the version is defined in one place per script.

gha-script/scanner-scripts/scancode_wheel_scan.sh
stutiibm: Done. Created a new script fetch_scancode_toolkit.sh (in scanner-scripts/) that clones scancode-toolkit once with up to 3 retries (with a 10s delay) to handle network issues, checks out the required version, and archives the source as scancode-toolkit-src.tar.gz. This script is called once in the install_scan_tools job and the archive is uploaded as a GHA artifact. All 5 wheel_build_pyXXX jobs then download this artifact and extract it directly — no redundant cloning. The pip install inside scancode_wheel_scan.sh still runs per job since the installed binary is not portable across runners. Verified working in GHA run — clone happens once, artifact downloaded and extracted successfully in all wheel build jobs.

cdeepali (approved):

PR #8437, opened 2026-07-19
CVE scanning enhancement

stutiibm:
## Checklist

### DIFF
--- a/gha-script/generalized_wheel_scanner.py
+++ b/gha-script/generalized_wheel_scanner.py
@@ -1,0 +1,1401 @@
+#!/usr/bin/env python3
+"""
+Generalized Wheel CVE Scanner
+Scans any Python wheel for CVEs including bundled .so files
+
+Usage:
+    python generalized_wheel_scanner.py <wheel_file.whl> [build_script.sh]
+
+    build_script.sh is optional — required for Lane 3 (source-built libs).
+    When provided, the scanner parses git clone/checkout lines to extract
+    source-built library names and versions, then queries NVD CPE + CVE APIs.
+"""
+
+import sys
+import json
+import os
+import re
+import subprocess
+import tempfile
+import time
+import urllib.request
+import urllib.parse
+import zipfile
+from pathlib import Path
+from datetime import datetime, timezone
+from typing import Dict, List, Optional, Set, Tuple
+import shutil
+
+
+NVD_API_BASE = "https://services.nvd.nist.gov/rest/json"
+
+
+class WheelCVEScanner:
+    """Generalized scanner for any Python wheel"""
+
+    def __init__(self, wheel_path: str, build_script_path: str = ""):
+        self.wheel_path = Path(wheel_path)
+        self.wheel_name = self.wheel_path.name
+        self.build_script_path = Path(build_script_path) if build_script_path else None
+        self.results = {
+            'wheel_file': self.wheel_name,
+            'scan_date': datetime.now(timezone.utc).isoformat(),
+            'steps': {}
+        }
+        
+    # Grype install URL
+    _GRYPE_INSTALL_URL = (
+        "https://raw.githubusercontent.com/anchore/grype/main/install.sh"
+    )
+
+    def _ensure_grype(self, retries: int = 3, retry_delay: int = 10) -> bool:
+        """
+        Ensure grype is available on PATH.  If not found, attempt to install it
+        via the official install script into /usr/local/bin.
+
+        Non-blocking — if all retries fail, prints a warning and returns False.
+        The caller skips grype-dependent steps but continues with Lane 3 (NVD).
+
+        Args:
+            retries:     number of install attempts (default 3)
+            retry_delay: seconds to wait between attempts (default 10)
+
+        Returns True if grype is available after this call, False otherwise.
+        """
+        if shutil.which('grype'):
+            print("  grype already installed.")
+            return True
+
+        print(f"  grype not found — attempting installation "
+              f"(up to {retries} tries, {retry_delay}s between retries)...")
+
+        for attempt in range(1, retries + 1):
+            try:
+                print(f"  Install attempt {attempt}/{retries}...")
+                result = subprocess.run(
+                    ['sh', '-c',
+                     f'curl -sSfL {self._GRYPE_INSTALL_URL} | sh -s -- -b /usr/local/bin'],
+                    capture_output=True, text=True, timeout=120
+                )
+                if result.returncode == 0 and shutil.which('grype'):
+                    print("  grype installed successfully.")
+                    return True
+                print(f"  Attempt {attempt} failed: {result.stderr.strip()[:200]}")
+            except Exception as e:
+                print(f"  Attempt {attempt} error: {e}")
+
+            if attempt < retries:
+                print(f"  Retrying in {retry_delay}s...")
+                time.sleep(retry_delay)
+
+        print("  WARNING: grype installation failed after all retries. "
+              "Lane 1, Lane 2 and Phase 1 (wheel direct scan) will be skipped. "
+              "Lane 3 (NVD) will still run.")
+        return False
+
+    def scan(self) -> Dict:
+        """Main scanning workflow"""
+        print(f"\n{'='*70}")
+        print(f"Scanning Wheel: {self.wheel_name}")
+        print(f"{'='*70}\n")
+
+        # Step 0: Ensure grype is available (install if needed, with retries)
+        print("Step 0: Checking grype availability...")
+        grype_available = self._ensure_grype()
+        self.results['grype_available'] = grype_available
+
+        # Step 1: Create system library inventory
+        print("\nStep 1: Creating system library inventory...")
+        self.results['steps']['inventory'] = self._create_system_inventory()
+
+        # Step 2: Extract wheel + Phase 1 grype scan on extracted wheel dir
+        print("\nStep 2: Extracting wheel...")
+        with tempfile.TemporaryDirectory() as tmpdir:
+            extract_dir = Path(tmpdir) / "wheel_contents"
+            self._extract_wheel(extract_dir)
+
+            # Step 2b: Phase 1 — grype scan directly on extracted wheel
+            print("\nStep 2b: Phase 1 — grype scan on extracted wheel...")
+            self.results['steps']['wheel_direct_scan'] = (
+                self._scan_wheel_direct(extract_dir) if grype_available
+                else {'skipped': True, 'reason': 'grype not available', 'matches': []}
+            )
+
+            # Step 3: Find bundled libraries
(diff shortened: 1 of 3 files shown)


gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+        self.wheel_name = self.wheel_path.name
+        self.build_script_path = Path(build_script_path) if build_script_path else None
+        self.results = {
+            'wheel_file': self.wheel_name,
+            'scan_date': datetime.now(timezone.utc).isoformat(),
+            'steps': {}
+        }
+        
+    # Grype install URL
+    _GRYPE_INSTALL_URL = (
```
cdeepali: Add a step in workflow to install grype, instead of installing it in the python script?

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+
+        Avoids [%{FILENAMES}] array iterator which fails on some RPM versions
+        when packages have mismatched array tag sizes.
+        """
+        packages = {}
+
+        try:
+            # Step 1: one call — all package names + versions
+            meta_result = subprocess.run(
+                ['rpm', '-qa', '--queryformat', '%{NAME}|%{VERSION}-%{RELEASE}\n'],
```
cdeepali: Can you try with, it will give package name, version and files all in one go. 

`rpm -qa --queryformat 'PKG:%{NAME}-%{VERSION}-%{RELEASE}.%{ARCH}\n[%{FILENAMES}\n]'`

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+            'timestamp': datetime.now(timezone.utc).isoformat(),
+            'packages': {}
+        }
+        
+        # Check if RPM-based system
+        if shutil.which('rpm'):
+            print("  Detected RPM-based system")
+            inventory['packages'] = self._get_rpm_packages()
+        # Check if DEB-based system
+        elif shutil.which('dpkg'):
```
cdeepali: Is there a need to prepare inventory for DEB based system, considering that the build happens on RHEL container?

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+        return packages
+    
+    def _get_dpkg_packages(self) -> Dict:
+        """Get library information from DEB packages"""
+        packages = {}
+        
+        try:
+            # Get all packages
+            result = subprocess.run(
+                ['dpkg-query', '-W', '-f=${Package}|${Version}\n'],
```
cdeepali: On Debian systems, dpkg stores the list of files for every installed package in plain text files under /var/lib/dpkg/info/<package>.list. Reading these files directly in Python is better compared to executing the dpkg-query and then dpkg for each package.

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+                for file_path in files_result.stdout.strip().split('\n'):
+                    if '.so' in file_path and Path(file_path).exists():
+                        try:
+                            so_files.append({
+                                'path': file_path,
+                                'realpath': str(Path(file_path).resolve()),
+                                'basename': Path(file_path).name
+                            })
+                        except:
+                            pass
```
cdeepali: Can use except Exception: here instead of passing silently.

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+        
+        for bundled_lib in bundled_libs:
+            bundled_name = bundled_lib['bundled_filename']
+            
+            # Extract base library name — strip auditwheel hash and all version info
+            # Works generically for any library naming convention:
+            #   libgfortran-2758a8fd.so.5.0.0    -> libgfortran
+            #   libopenblasp-r0-a8f83a82.3.32.so -> libopenblasp-r0
+            #   libabsl_base-eb206faa.so.2401.0.0 -> libabsl_base
+            base_name = re.sub(r'-[a-f0-9]{8,}', '', bundled_name)  # Remove hex hash
```
cdeepali: Can this be replaced with single regex?

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+            print(f"  {len(result['unknown'])} deps not found in site-packages: {result['unknown']}")
+
+        return result
+
+    def _find_site_packages(self) -> List[Path]:
+        """Find all site-packages directories on the system"""
+        site_dirs = []
+        try:
+            result = subprocess.run(
+                ['find', '/', '-name', 'site-packages', '-type', 'd',
```
cdeepali: Can the following be used instead?

```
import site
dirs = site.getsitepackages() + [site.getusersitepackages()] 
```

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+            return result
+
+        print(f"  Found site-packages dirs: {[str(d) for d in site_packages_dirs]}")
+
+        # Recursive walk
+        visited: Set[str] = set()
+        queue = [wheel_pkg_name]
+
+        while queue:
+            pkg = queue.pop(0)
```
cdeepali: use deque instead.

cdeepali: use deque instead.

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+                print(f"  Install attempt {attempt}/{retries}...")
+                result = subprocess.run(
+                    ['sh', '-c',
+                     f'curl -sSfL {self._GRYPE_INSTALL_URL} | sh -s -- -b /usr/local/bin'],
+                    capture_output=True, text=True, timeout=120
+                )
+                if result.returncode == 0 and shutil.which('grype'):
+                    print("  grype installed successfully.")
+                    return True
+                print(f"  Attempt {attempt} failed: {result.stderr.strip()[:200]}")
```
cdeepali: Use constant instead of magic number 200.

gha-script/generalized_wheel_scanner.py
```diff
@@ -0,0 +1,1401 @@
...
+        """Single NVD API GET call. Reads NVD_API_KEY from environment."""
+        req = urllib.request.Request(url)
+        api_key = os.environ.get("NVD_API_KEY", "")
+        if api_key:
+            req.add_header("apiKey", api_key)
+        try:
+            with urllib.request.urlopen(req, timeout=30) as resp:
+                return json.loads(resp.read())
+        except Exception as e:
+            print(f"  Warning: NVD API call failed ({url}): {e}")
```
cdeepali, line 827: Check that this does not print the Key here.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. Removed _ensure_grype() method and _GRYPE_INSTALL_URL entirely from the Python script. Grype is now installed once in a dedicated install_scan_tools workflow job using wget with checksum verification, cached as a scan-tools-cache artifact, and downloaded by each wheel_build_pyXXX job. The Python script now simply checks shutil.which('grype') and skips grype-dependent lanes gracefully if not found.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. Replaced the two-call approach with a single rpm -qa --queryformat 'PKG:%{NAME}|%{VERSION}-%{RELEASE}\n[%{FILENAMES}\n]' call as suggested. Lines starting with PKG: are parsed as package headers; subsequent lines are file paths belonging to that package.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. Removed the elif shutil.which('dpkg') branch from _create_system_inventory() and deleted _get_dpkg_packages() entirely. The builds run on RHEL containers so only RPM-based inventory is needed.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed as part of Comment 3 — _get_dpkg_packages() has been removed entirely since DEB inventory is not needed for RHEL-based builds.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. The bare except: pass was inside _get_dpkg_packages() which has been removed entirely (Comment 3). All remaining exception handlers in the file already use except Exception as e:.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. Replaced three chained re.sub() calls with a single regex: re.sub(r'(-[a-f0-9]{8,})?(\.\d[^/]*)?\.so.*$', '', bundled_name) which strips the auditwheel hash, version suffix, and .so extension in one pass.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. Replaced the find / subprocess in _find_site_packages() with the suggested approach: site.getsitepackages() + [site.getusersitepackages()] using Python's built-in site module. Added import site at the top of the file.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. Added from collections import deque import, changed queue = [wheel_pkg_name] to queue: deque = deque([wheel_pkg_name]), and queue.pop(0) to queue.popleft() for O(1) dequeue instead of O(n) list shift.

stutiibm: Addressed. Added from collections import deque import, changed queue = [wheel_pkg_name] to queue: deque = deque([wheel_pkg_name]), and queue.pop(0) to queue.popleft() for O(1) dequeue instead of O(n) list shift.

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. Added module-level constant MAX_STDERR_LOG_CHARS = 200 and replaced all [:200] slices throughout the file with [:MAX_STDERR_LOG_CHARS].

gha-script/generalized_wheel_scanner.py
stutiibm: Addressed. Confirmed and documented: the API key is read from os.environ.get("NVD_API_KEY", "") and passed only via req.add_header("apiKey", api_key) as an HTTP request header. It is never interpolated into any print() or log statement. Added an explicit comment in the code confirming this.

cdeepali (approved):

PR #8518, opened 2026-07-30
fix: normalize TESTED_ON value in execute_changed_scripts.py to match build_package.sh patterns

stutiibm:
## Checklist

### DIFF
--- a/gha-script/execute_changed_scripts.py
+++ b/gha-script/execute_changed_scripts.py
@@ -92,13 +92,13 @@ def main():
 
         # Check if script exists
         if not os.path.exists(script_path):
-            print(f"⚠️  Script not found: {script_path}")
+            print(f"  Script not found: {script_path}")
             continue
 
         # Extract metadata from script header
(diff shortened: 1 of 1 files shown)


stutiibm:
execute_changed_scripts.py extracted the "Tested on" header value with a
plain .strip(), preserving internal spaces (e.g. "UBI 10.2").

build_package.sh pattern matches only UBI:10* and UBI10* — neither matches
"UBI 10.2" (space between UBI and 10), so all UBI 10 build jobs fell through
to the else branch and pulled the UBI 8.7 image instead.

read_buildinfo.sh already handles this correctly via:
  tr -d '[:space:]' | tr '[:lower:]' '[:upper:]'

Apply the same normalization in execute_changed_scripts.py for the
tested_on field only: strip spaces and uppercase the value so
"UBI 10.2" → "UBI10.2", which correctly matches the UBI10* pattern
in build_package.sh.

cdeepali (approved):

PR #8521, opened 2026-07-30
fix: use SUDO variable for privileged commands in create_wheel_wrapper.sh

stutiibm:
## Checklist

### DIFF
--- a/gha-script/create_wheel_wrapper.sh
+++ b/gha-script/create_wheel_wrapper.sh
@@ -7,13 +7,15 @@ EXTRA_ARGS=${3:-""}
 POST_PROCESS_SCRIPT_PATH=${4:-"post_process_wheel.py"}
 CURRENT_DIR=$(pwd)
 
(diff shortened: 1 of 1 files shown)


cdeepali (approved):

stutiibm:
The previous fix hardcoded 'sudo' for tee and ldconfig calls in the
Python source-build cases (3.10, 3.13, 3.14), but sudo is not installed
in a standard root UBI container — only added by dockerfile_non_root.
This caused root container builds to fail with:
  sudo: command not found
  libpython3.13.so.1.0: cannot open shared object file

Introduced a SUDO variable alongside the existing YUM variable, set
at the top of the script using the same id -u check:
  non-root: SUDO="sudo"
  root:     SUDO=""

Replaced all three hardcoded 'sudo tee' and 'sudo ldconfig' occurrences
with '$SUDO tee' and '$SUDO ldconfig', making them work correctly for
both root and non-root container builds.

PR #8533, opened 2026-08-03
Ubi10 pipeline changes

stutiibm:
## Checklist


stutiibm:
## Summary of Changes

This PR adds full UBI10 support to the CI pipeline and fixes several routing
bugs that caused wrong build scripts to run across all UBI versions.

---

### gha-script/read_buildinfo.sh

- Added `read_tested_on()` helper to normalise `# Tested on` values from build
  scripts (strips whitespace, uppercases, collapses `UBI : 9.3` -> `UBI:9.3`).
- Added full UBI10 bucketing: `SCRIPT_UBI8`, `SCRIPT_UBI9`, `SCRIPT_UBI10` are
  now exported as `{"script":"...","tested_on":"..."}` JSON objects.
- `build_script` in `build_info.json` can now be a list as well as a string;
  both forms are iterated and each entry is bucketed by its own `# Tested on`.
- Removed a stray Unicode checkmark character from a comment.

### gha-script/build_package.sh

- Replaced the fragile UBI8/UBI9/UBI10 if-elif-else chain with a generalised
  `resolve_docker_image()` function.
- Normalises all TESTED_ON formats (`UBI:9.3`, `UBI 9.3`, `ubi9.3`,
  `UBI: 9.3`, `UBI 10`) before extracting the major version integer.
- Supports UBI8, UBI9, and UBI10+ by construction — adding UBI11 in the future
  requires no code change.
- Falls back to sensible defaults when only a major version (no minor) is given.

### gha-script/build_wheels.sh

- Applied the same `resolve_docker_image()` generalisation as `build_package.sh`.
- Removed the hard-coded UBI10/Python-3.10-3.11 early-exit guard; the function
  handles unsupported combinations cleanly.

### gha-script/build_wheels.py

- Passes `GRYPE_BIN` and `ENABLE_CVE_SCAN` environment variables into the wheel
  build container so the pre-cached grype binary and the PR-build skip flag are
  visible inside the container.

### gha-script/create_wheel_wrapper.sh

- Added UBI10 support: selects `gcc-toolset-15` for UBI >= 10, activates it via
  `PATH` prepend (SCL `enable` scripts were dropped in UBI10).
- Simplified privilege handling: the wrapper now always runs as root inside the
  container, so the `$YUM`/`$SUDO` indirection was removed.

### gha-script/generalized_wheel_scanner.py

- Resolves the grype binary via `GRYPE_BIN` env var first (container-injected
  path), falling back to `shutil.which('grype')` for local runs.
- Both `_scan_wheel_direct` and `_scan_for_cves` now call `self.grype_bin`
  instead of the bare string `'grype'`.

### gha-script/post_process_wheel.py

- Changed `os.environ["key"]` (hard crash on KeyError) to `os.environ.get()`
  for the two COS credential variables.
- When credentials are absent (PR build environment), the script exits with
  code 0 instead of raising an unhandled exception, so PR wheel jobs do not
  fail post-processing.

### gha-script/scanner-scripts/grype_code_scan.sh

- Removed runtime `wget` + install of grype; now uses the pre-cached binary
  supplied via `$GRYPE_BIN`.
- Fails fast with a clear error if `GRYPE_BIN` is not set.

### gha-script/scanner-scripts/grype_image_scan.sh

- Same change as `grype_code_scan.sh`: use `$GRYPE_BIN` instead of downloading
  grype at runtime.

---

### .github/workflows/pr-build.yaml

**Routing fixes**

- `build_info` job now emits three per-UBI-major outputs: `script_ubi8`,
  `script_ubi9`, `script_ubi10` — each a `{"script":"...","tested_on":"..."}`
  JSON object or an empty string.
- Version block selection uses a best-match count loop over all changed `.sh`
  files instead of the previous break-on-first-match logic, so the version
  block with the most overlap wins. The wildcard `"*"` key is excluded from
  this search.
- Safety-net step: after `read_buildinfo.sh`, any UBI slot still empty is
  filled directly from the changed script's own `# Tested on` header, ensuring
  every changed script in the PR is tested regardless of `build_info.json`
  structure.
- Each `build_ubiN` and `wheel_build_ubiN_pyXXX` job now captures
  `THIS_SCRIPT`/`THIS_TESTED_ON` before sourcing `variable.sh` and re-exports
  them after, preventing `variable.sh` from overwriting the correct per-UBI
  values.

**New jobs**

- `build_ubi10`: runs `execute_changed_scripts.py` inside a UBI10 container
  when a UBI10 script is present.
- UBI8 wheel jobs: `py311`, `py312` (py310, py313, py314 removed — not
  supported on UBI8).
- UBI9 wheel jobs: `py310`, `py311`, `py312`, `py313`, `py314`.
- UBI10 wheel jobs: `py312`, `py313`, `py314` (py310/py311 not available on
  UBI10).

---

### .github/workflows/currency-build.yaml

**New jobs**

- `build_ubi10`: currency build inside a UBI10 container.
- `source_scanner_ubi10`: trivy/syft/grype source scan after `build_ubi10`.
- `wheel_license_extract_ubi10`: aggregates scancode results for UBI10 wheels.
- UBI8 wheel jobs: `py311`, `py312` (py310, py313, py314 removed).
- UBI9 wheel jobs: `py310`, `py311`, `py312`, `py313`, `py314`.
- UBI10 wheel jobs: `py312`, `py313`, `py314`.

**Grype caching**

- `install_scan_tools` job downloads grype once and uploads it as the
  `scan-tools-cache` artifact; all scan jobs download it from there instead of
  fetching it at runtime. Checksum verification is performed before extraction.

**Bug fixes**

- All wheel jobs now re-export `BUILD_SCRIPT` and `TESTED_ON` after
  `source variable.sh` to prevent the backward-compat first-script value from
  overwriting the correct per-UBI values.
- `wheel_license_extract_ubi8` `needs` list and `if` condition updated to match
  the trimmed UBI8 wheel job set (`py311`, `py312` only).

.github/workflows/currency-build.yaml
```diff
@@ -67,15 +70,19 @@ jobs:
...
       ENABLE_SYFT: ${{ inputs.enable_syft }}
       ENABLE_GRYPE: ${{ inputs.enable_grype }}
+    outputs:
+      script_ubi8:  ${{ steps.emit.outputs.script_ubi8 }}
+      script_ubi9:  ${{ steps.emit.outputs.script_ubi9 }}
+      script_ubi10: ${{ steps.emit.outputs.script_ubi10 }}
     steps:
       - name: Checkout code
-        uses: actions/checkout@v6
+        uses: actions/checkout@v4
```
cdeepali, line 79: Is this change intentional?

cdeepali (changes requested):

gha-script/create_wheel_wrapper.sh
```diff
@@ -7,28 +7,20 @@ EXTRA_ARGS=${3:-""}
...
-    SUDO=""
-fi
+# install git  -  required by generate_sha() for all Python versions and UBI versions
+yum install -y git
 
-# install gcc — select toolset version based on UBI major version
+# install gcc  -  select toolset version based on UBI major version
 UBI_MAJOR=$(grep -oP '(?<=^VERSION_ID=")[0-9]+' /etc/os-release || grep -oP 'release \K[0-9]+' /etc/redhat-release 2>/dev/null || echo "8")
 if [[ "$UBI_MAJOR" -ge 10 ]]; then
     GCC_TOOLSET="gcc-toolset-15"
```
cdeepali, line 16: Can we set GCC_TOOLSET in the if/else block and move the code to install and enable GCC_TOOLSET outside of if.

gha-script/post_process_wheel.py
```diff
@@ -43,9 +43,15 @@
...
-COS_SERVICE_INSTANCE_ID = os.environ["GHA_CURRENCY_SERVICE_ID"]
+# COS configuration
+# These are only available in currency builds (not PR builds).
+# If missing, post-processing is skipped gracefully.
+COS_API_KEY = os.environ.get("GHA_CURRENCY_SERVICE_ID_API_KEY", "")
+COS_SERVICE_INSTANCE_ID = os.environ.get("GHA_CURRENCY_SERVICE_ID", "")
+
+if not COS_API_KEY or not COS_SERVICE_INSTANCE_ID:
+    logger.info("COS credentials not set - skipping post-processing (PR build environment)")
+    sys.exit(0)
```
cdeepali, line 54: Should this fail here ? as post-processing is integral part of the workflow

.github/workflows/currency-build.yaml
stutiibm: yes, v4 is stable version where as v6 is latest version.
So better to go with stable version

gha-script/create_wheel_wrapper.sh
stutiibm: We can do that but again we need to add one more if-else block to export path based on the gcc_toolset installed version as both of gcc_toolset version follows different ways of path export

gha-script/post_process_wheel.py
stutiibm: I've reworked the fix.

Instead of silently swallowing missing credentials in post_process_wheel.py,
the guard is now in create_wheel_wrapper.sh — the same ENABLE_CVE_SCAN=false
flag that pr-build.yaml already sets on every PR wheel job is used to skip the
post_process_wheel.py call entirely before it is even invoked.

post_process_wheel.py is reverted to the original hard-fail (os.environ["key"]),
so if credentials are ever absent in a currency build it will fail immediately
and loudly rather than silently succeed. PR builds never reach post_process_wheel.py
at all, which is the correct behaviour since PRs only validate that the wheel
can be built, not that it can be published.

gha-script/create_wheel_wrapper.sh
cdeepali: If same comand does not work, then need not change the if else block .

cdeepali (approved):

PR #8573, opened 2026-08-06
fix: use build_info.json .version field as VERSION in wheel and build…

stutiibm:
… jobs

VERSION was being overwritten by the version block key selection logic (e.g. "v2.10.*") which is a wildcard pattern used only for matching the correct config block in build_info.json — not a real git tag. This caused wheel jobs to pass the pattern as the version argument to build_wheels.sh, which then failed inside the container with:

  git checkout 'v2.10.*'
  error: pathspec 'v2.10.*' did not match any file(s) known to git

Fix: capture the raw .version field from build_info.json into PACKAGE_VERSION immediately after it is read, before VERSION is potentially overwritten by the block key selection. Write PACKAGE_VERSION (the real version, e.g. "v2.11.0") into scanner-env.sh instead of VERSION, so all downstream jobs (wheel builds) receive a valid git tag.

VERSION (the block key) is preserved unchanged for read_buildinfo.sh, which requires it for version block matching. No other behaviour changes — multi-UBI routing, build jobs, and all control flags are unaffected.

## Checklist

### DIFF
--- a/.github/workflows/pr-build.yaml
+++ b/.github/workflows/pr-build.yaml
@@ -179,6 +179,10 @@ jobs:
 
           PACKAGE_NAME=$(jq -r '.package_name // ""' $BUILD_INFO_FILE)
           VERSION=$(jq -r '.version // ""' $BUILD_INFO_FILE)
+          # Preserve the raw .version field from build_info.json as the real package
+          # version (e.g. "v2.11.0"). VERSION may be overwritten below by the matched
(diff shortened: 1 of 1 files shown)


stutiibm:
Tested change with dummy PR as below :

Issue PR : https://github.com/ppc64le/build-scripts/actions/runs/31070248541?pr=8567
Tested PR : https://github.com/stutiibm/build-scripts/actions/runs/31079190194

cc : @cdeepali

cdeepali (approved):

PR #8679, opened 2026-08-27
fix(scipy): support v1.18.1 and future versions via wildcard fallback

pjsharath28:
### Changes
- Add `"*"` wildcard entry in `build_info.json` pointing to `scipy_v1.18.0_ubi_9.6.sh` so any version without an explicit key resolves to the latest build script
- Fix version assertion in `scipy_v1.18.0_ubi_9.6.sh`: remove quotes from the heredoc opening delimiter (`<< 'PYEOF'` → `<< PYEOF`) so the shell expands `${PACKAGE_VERSION#v}` dynamically before passing it to Python

### Root Cause
Currency build for `v1.18.1` was failing because:
1. No `"*"` wildcard fallback in `build_info.json` (silently fell back to top-level `build_script`, which happened to be correct but is implicit)
2. The heredoc used `<< 'PYEOF'` (single-quoted), which suppresses all shell expansion inside the block — `${PACKAGE_VERSION#v}` was passed as a literal string to Python instead of the actual version value, causing the assertion to always fail for any version other than `1.18.0`

---

## Checklist


pjsharath28:
Created PR from incorrect branch, unable to update the source branch. Will be closing this PR and creating a new one.

PR #8692, opened 2026-09-01
logstash: Added OSS build script, patch, Dockerfile for v9.4.5

prachi-gaonkar:
## Checklist

### DIFF
--- a/l/logstash/Dockerfiles/9.4.5_ubi_9.8/Dockerfile
+++ b/l/logstash/Dockerfiles/9.4.5_ubi_9.8/Dockerfile
@@ -1,0 +1,160 @@
+# syntax=docker/dockerfile:1
+FROM registry.access.redhat.com/ubi9/ubi:9.8 AS build
+
+ARG PACKAGE_VERSION=v9.4.5
+# PATCH is pinned to v9.4.5 intentionally; newer versions may not have a patch available
+ARG PATCH=https://raw.githubusercontent.com/ppc64le/build-scripts/master/l/logstash/logstash_v9.4.5.patch
+ENV PACKAGE_NAME=logstash
+
+RUN yum -y update && yum install -y \
+    git procps yum-utils wget ncurses make gcc-c++ libffi-devel \
+    java-21-openjdk java-21-openjdk-devel java-21-openjdk-headless
+
+# bison and readline-devel are not in UBI9; add CentOS Stream 9 repos
+RUN rpm --import https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official && \
+    printf '[cs9-appstream]\nname=CentOS Stream 9 - AppStream\nbaseurl=https://mirror.stream.centos.org/9-stream/AppStream/ppc64le/os/\ngpgcheck=1\ngpgkey=https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official\nenabled=1\n' \
+        > /etc/yum.repos.d/cs9-appstream.repo && \
+    printf '[cs9-baseos]\nname=CentOS Stream 9 - BaseOS\nbaseurl=https://mirror.stream.centos.org/9-stream/BaseOS/ppc64le/os/\ngpgcheck=1\ngpgkey=https://www.centos.org/keys/RPM-GPG-KEY-CentOS-Official\nenabled=1\n' \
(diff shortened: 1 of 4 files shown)


l/logstash/logstash_v9.4.5.patch
```diff
@@ -0,0 +1,197 @@
...
+                 return "x64"
+                 break
++            case "ppc64le":
++            case "ppc64":
++            return "ppc64le"
+             case "arm64":
+                 return "aarch64"
+                 break
+@@ -1035,15 +1056,19 @@ def getTargetPlatforms() {
+     String arch = selectArch()
```
Amitkumar282, line 76: Since ppc64le is already handled in the else if above, the else branch should retain the existing ARM64 behavior. With the current change, ARM64 (and any other non-x86/non-ppc64le architecture) will incorrectly resolve to ppc64le, and the existing ARM64 return becomes unreachable.

l/logstash/Dockerfiles/9.4.5_ubi_9.8/Dockerfile
```diff
@@ -0,0 +1,161 @@
+# syntax=docker/dockerfile:1
+# Stage 1: build the Logstash OSS tarball for ppc64le
+FROM registry.access.redhat.com/ubi9/ubi:latest AS build
+
+ARG PACKAGE_VERSION=v9.4.5
+ARG PATCH=https://raw.githubusercontent.com/ppc64le/build-scripts/master/l/logstash/logstash_v9.4.5.patch
```
Amitkumar282: PATCH is hardcoded to the v9.4.5 patch while PACKAGE_VERSION is exposed as a build argument. If PACKAGE_VERSION is overridden, we could end up applying the wrong patch.

l/logstash/logstash_v9.4.5.patch
prachi-gaonkar: Thank you for pointing it out. 'ppc64le' is now handled separately, and the existing ARM64 fallback has been preserved.

l/logstash/logstash_oss_9.4.5_ubi9.8.sh
```diff
@@ -0,0 +1,220 @@
...
+
+if ! ./gradlew assemble; then
+    echo "------------------${PACKAGE_NAME}:assemble_fails-------------------------------------"
+    echo "${PACKAGE_NAME} | ${PACKAGE_URL} | ${PACKAGE_VERSION} | GitHub | Fail | Assemble_Fails"
+    exit 1
+fi
+echo "------------------${PACKAGE_NAME}:assemble_success-------------------------------------"
+
+# ── 7. Build the OSS tar ──────────────────────────────────────────────────────
+# Run as root before the test-user chown, so no ownership restore is needed.
```
Amitkumar282: Can we shorten the comments?

l/logstash/logstash_v9.4.5.patch
```diff
@@ -0,0 +1,188 @@
...
+ }
+diff --git a/logstash-core/lib/logstash/agent.rb b/logstash-core/lib/logstash/agent.rb
+index c16b47ec9..901d7f945 100644
+--- a/logstash-core/lib/logstash/agent.rb
++++ b/logstash-core/lib/logstash/agent.rb
+@@ -685,7 +685,7 @@ class LogStash::Agent
+       database_metric = LogStash::GeoipDatabaseManagement::Metric.new(geoip_metric)
+       database_manager = LogStash::GeoipDatabaseManagement::Manager.instance
+       database_manager.database_metric = database_metric
+-    rescue LoadError => e
```
Amitkumar282: If `LoadError` occurs, the current trace message makes sense and tells us that the `DatabaseManager` is not in the classpath. But if an `ArgumentError` occurs, the same trace message could be misleading and create confusion about the actual issue.

Can we handle these separately with an appropriate message for each? For example:


rescue LoadError => e
  @logger.trace("DatabaseManager is not in classpath", exception: e.message, backtrace: e.backtrace)

rescue ArgumentError => e
  @logger.trace("Failed to initialize DatabaseManager due to ArgumentError", exception: e.message, backtrace: e.backtrace)

Sapana-Khemkar (approved):

Sapana-Khemkar:
Prachi to verify CI once merged.

